PR #15, opened 2017-05-19
Added travis.yml

sanikakulkarni:


sanikakulkarni:
The job has to be edited to include the correct file paths once the wf_tests are added into the repo

PR #24, opened 2017-06-13
WAR-1170: Changing <IDF> tag to <InputDatFile> in the Suite Page

sanikakulkarni:


kaheichan:
Checked that katana now create InputDataFile tag and also overwrite wrong IDF tag to InputDataFile

PR #25, opened 2017-06-13
Feature/war 890 Add option to escape ANSI sequence

kaheichan:
Add an optional tag/attrib escape which takes "yes" or "true" to filter out ANSI sequence on terminal
For example in the prompt_fix.zip, we had
[0m[01;34mCliTestDir2[0m [01;34mDocuments[0m [01;34mMusic[0m [01;34mPublic[0m [01;34mVideos[0m
[01;34mDesktop[0m [01;34mDownloads[0m [01;34mPictures[0m [01;34mTemplates[0m
]0;username@hostname: \~username@hostname:\~$
before

but after filtering out all the escape sequence, we have
CliTestDir2 Documents Music Public Videos
Desktop Downloads Pictures Templates
username@hostname:~$

which increase readability and helps eliminate regex catching wrong text
(Regex may match the escape sequence which doesn't show in normal terminal and affect test status)


warrior/Framework/ClassUtils/warrior_connect_class.py
```diff
@@ -415,7 +416,10 @@ def connect_ssh(self):
         if WarriorCliClass.cmdprint:
             pNote("connectSSH: :CMD: %s" % command)
             return None, ""
-        child = self.pexpect.spawn(command, timeout=int(self.timeout))
```
satyakrish: lines 419-422 are repeated in several places in this PR, is it possible to convert it to a generic function in cli_utils

PR #26, opened 2017-06-14
war 897: As a user i need the ability to make nested ssh connections

venkat1077:
1. Added support for nested SSH connections using paramiko.
2. Converted connect_ssh, connect_telent, send_command, disconnect_telnet & disconnect util functions as class methods.
3. Added warning message in util methods to let the users to know that these methods are obsolete & will be deprecated soon.

Use-case for nested connection : Connecting target system via intermediate system from the execution system. Here the target is not directly reachable from execution system but can be reached via intermediate system.


kaheichan:
@venkat1077 Small change in pylint please

venkat1077:
@kaheichan I have addressed pylint failure and also added warning message in 'to be deprecated' functions.

PR #29, opened 2017-06-15
WAR-1037 : Execution summary is not getting printed correctly at the end of warrior execution

Annam16:
Execution summary is not getting printed correctly at the end of warrior execution if we have the same suite "n" number of times in a project file


kaheichan:
Please add the ticket number, thank you

bjrr:
@Annam16 Please correct the subject to follow the use convention. WAR-123: Subject text

PR #30, opened 2017-06-15
Feature/WAR-725 automate the testing of iterative test suite, parallel/sequential ts with mix par/seq tc

kaheichan:
Since both WAR-947 and WAR-725 uses the same set of keywords, I have to branch one from the other...
The test are done in the following way:
For iterative testsuite, have 3 systems in the datafile with filename
the testsuite read the datafile and execute per system

For both tests, it use the following mechanism
in sequential case, it creates file 1 by 1, verify individual file exist, when finish, do a count (should have only 1 file) and delete
in parallel case, it creates all file together, verify individual file exist, do a count for 3 files and delete

When all 4 comb of par/seq ts/tc present, it creates a bug that hang execution which is reported in WAR-1153
so the pj file for parallel execution is separate

To test and verify, please run
pj_iterative_execution.xml
pj_parallel_execution.xml
pj_parallel_execution_2.xml


kaheichan:
Fixed is delivered in #93

PR #35, opened 2017-06-19
Bugfix/war 1195: verify_on doesn't work properly in test data

venkat1077:
Issue: 'verify_on' option in td is not working as expected when the system name given at the keyword level is different from the 'sys & verify_on' tag values in td block.
Example:
    system name given at step level - system1
    sys tag value in td block - system2
    verify_on value in td block - system2
Issue behavior: 'send' command will be executed in system2 and the response will be collected from system1.
Expected behavior: 'send' command should be executed in system2 and the verification response should be collected from the system2
Note - This wouldn't occur when the 'verify_on' tag is not given or left empty

Reason - We are creating a thread for collecting log from verify_on system during command verification process since 'verify_on' value is different from the step level system name, but this would give nothing since the logs were already collected during command execution process because of the sys tag.

Solution - When the sys tag value is given, we have to treat it as step level system name for execution & verification process.


venkat1077:
@kaheichan Fixed pylint issue

kaheichan:
@venkat1077 Sorry just saw this, approved

PR #36, opened 2017-06-19
Bugfix/WAR-1197 Katana : Insert New Step hides the first saved Step

keenan35i:


kaheichan:
Please change branch from keenan35i-patch-1 to bugfix/WAR-1197 please

PR #38, opened 2017-06-20
bugfix/WAR-1203: Unable to use the environment variable with dots in it

venkat1077:
Issue: Unable to use the environment variable with dots in it
Reason: For ENV substitution, 'get_var_by_string_prefix' method just uses the second part of a string after splitting it using '.'(dot).
Solution:  Use maxsplit argument to get the string comes after first occurrence of '.'(dot).

### DIFF
--- a/warrior/Framework/Utils/data_Utils.py
+++ b/warrior/Framework/Utils/data_Utils.py
@@ -1338,9 +1338,9 @@ def get_filepath_from_system(datafile, system_name, *args):
 
 
 def get_var_by_string_prefix(string):
-    if string.startswith("ENV"):
+    if string.startswith("ENV."):
         return os.environ[string.split('.', 1)[1]]
-    if string.startswith("REPO"):
+    if string.startswith("REPO."):
         keys = string.split('.')
         val = get_object_from_datarepository(keys[1])
         for key in keys[2:]:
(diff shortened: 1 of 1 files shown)


warrior/Framework/Utils/data_Utils.py
```diff
@@ -1339,7 +1339,7 @@ def get_filepath_from_system(datafile, system_name, *args):
 
 def get_var_by_string_prefix(string):
     if "ENV" in string:
```
warriorcquser: change this condition and action to if string starts with ENV. ten remove the 'ENV.'

satyakrish: change this condition and action to if string starts with 'ENV.' remove the  first 4 chars and use the rest to fetch the environment variable

warrior/Framework/Utils/data_Utils.py
```diff
@@ -1339,7 +1339,7 @@ def get_filepath_from_system(datafile, system_name, *args):
...
-        return os.environ[string.split('.')[1]]
+        return os.environ[string.split('.', 1)[1]]
     if "REPO" in string:
```
satyakrish: change this condition and action to if string starts with 'REPO.' remove the first 5 chars and use the rest to fetch the value from data repository.

warrior/Framework/Utils/data_Utils.py
venkat1077: Modified condition to check if the string starts with 'ENV', remaining will work fine.

warrior/Framework/Utils/data_Utils.py
venkat1077: Modified condition to check if the string starts with 'REPO', remaining will work fine.

PR #39, opened 2017-06-20
bugfix/WAR-1193 Katana : Testcase Screen - 'Goto' not showing up its input field.

keenan35i:


kaheichan:
Verified that the goto input field for step is showing up on testcase screen for both existing testcase and new testcase

PR #45, opened 2017-06-25
war 1074 : selenium functionalities in headless mode

Annam16:
To run the selenium functionality tests in headless mode
Need to install pyvirtualdisplay and Xvfb to run run in headless mode
If it is not installed error msg would be thrown to the user and it would execute the tcs normally.


warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -175,10 +185,22 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
                 browser_details = selenium_Utils.\
                     get_browser_details(browser, self.datafile, **arguments)
             if browser_details is not None:
+                try:
+                    from pyvirtualdisplay import Display
+                    display = Display(visible=0, size=(1024, 768))
+                    display.start()
+                    pNote("Running in headless mode")
```
navvrnd: Why pNote? Can we use print_info instead?

Annam16: @navvrnd  : "pNote" default "print_type" value is "print_info". So I have used that.

sanikakulkarni: Hi @Annam16,

Is there a way for the user to specify if they want to run the testcase in headless mode? We can add an argument to the browser_launch keyword called headless_mode, which by default be set to False. The user can then set it True to run the testcases in headless mode.

The current implementation defaults all runs to headless mode and there seems to be no way around that barring uninstalling the headless mode libraries.

Let me know what you think.

warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -45,6 +45,16 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
...
 
+        If the user wants to test the selenium functionalities in headless
+        mode, install pyvirtualdisplay and Xvfb. To install it follow
+        the below steps.
+
+        1. To install pyvirtualdisplay:
+               pip install pyvirtualdisplay
+
+        2. To install Xvfb:
+               sudo apt-get install Xvfb
```
sanikakulkarni: Please also add a comment that shows how to install Xvfb on RedHat based systems (yum install)

warrior/Actions/SeleniumActions/browser_actions.py
Annam16: yes @sanikakulkarni . I will add these changes.

warrior/Actions/SeleniumActions/browser_actions.py
Annam16: Done @sanikakulkarni

warrior/Actions/SeleniumActions/browser_actions.py
Annam16: added a new argument "headless_mode" @sanikakulkarni . Input would be "yes" or "no".

sanikakulkarni:
Hi @Annam16,

Can you please merge develop into your branch?

warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -179,10 +200,23 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
...
+                        from pyvirtualdisplay import Display
+                        display = Display(visible=0, size=(1024, 768))
+                        display.start()
+                        pNote("Running in headless mode")
+                    except ImportError:
+                        print_error("pyvirtualdisplay is not installed in order "
+                                    "to launch the browser in headless mode")
+                    except:
+                        print_error("Xvfb is not installed in order "
+                                    "to launch the browser in headless mode")
```
sanikakulkarni: Hi @Annam16,

If any of the exceptions mentioned above occur - the step should fail. Today, if I do not have pyvirtualdisplay/Xvfb installed on my system, it prints outs this exception, but eventually my step passes (because I do not really need the headless mode on my ubuntu); but the step should fail indicating that something went wrong.

warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -179,10 +200,23 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
                 browser_details = selenium_Utils.\
                     get_browser_details(browser, datafile=self.datafile, **arguments)
             if browser_details is not None:
+                if str(headless_mode).strip().lower() == "yes":
```
sanikakulkarni: Hi @Annam16,

Instead of a direct ```==```, can we have a ReGex that supports "y" and "yes" (case insensitive)? Please also update the docstrings accordingly.

Annam16:
@sanikakulkarni 
addressed all the review comments. Please review and approve.

Annam16:
Hi @sanikakulkarni Please review.

sanikakulkarni:
Hi @brettleehari,

I have a couple of comments in the PR for WAR-1074 that is going into out internal Bitbucket repository. Please address those.

brettleehari:
@sanikakulkarni Comments addressed in bitbucket.

bjrr:
@brettleehari Please resolve merge conflicts

brettleehari:
This conflict resolution requires change in logic

brettleehari:
@sanikakulkarni, 
                              Can you review. Discussed with the developer (external contributor) whose code was in conflict with this pull request. We identified and changed the logic so that both the functionality is available to the user. Headless mode and selenium 3.0 support.

bjrr:
Hi @brettleehari,

Please run the selenium regression (twice - once for selenium 2 and once for selenium 3) and attach the logs to the jira ticket.

bjrr:
Hi @brettleehari,

Adding a ```<headless_mode>yes</headless_mode>``` tag in the data file has no effect on the testcase. Adding ```<argument name="headless_mode" value="yes"/>``` in the testcase works as expected. Please correct this.

warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -215,16 +237,32 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
...
+                        from pyvirtualdisplay import Display
+                        display = Display(visible=0, size=(1024, 768))
+                        display.start()
+                        pNote("Running in headless mode")
+                    except ImportError:
+                        print_error("pyvirtualdisplay is not installed in order "
+                                    "to launch the browser in headless mode")
+                        status = False
+                    except:
+                        print_error("Xvfb is not installed in order "
```
bjrr: Hi @brettleehari,

This line will cal every exception that occurs and will print out the statement on line 255 - which can result in a false exception detection. Please change lines 254, 255, and 256 to:

```
except Exception as e:
     print_error("An Exception Occurred: {0}".format(e))
```

brettleehari: Assumption is line 246 would be caught by Import Error exception, line 247 & 248 would mostly hit exception when xvfb is not installed this is caught by the general exception :). line 249 exception is very unlikely but I agree with you comment that line 249 case makes this comment more solid. 
Thanks and changed.

brettleehari:
@bjrr The mentioned issue works as expected. When you give headless mode in data file <headless_mode>yes</headless_mode> it works. 
Can you please check again and update the comment appropriately. 

Below is your comment.
`Adding a <headless_mode>yes</headless_mode> tag in the data file has no effect on the testcase. Adding <argument name="headless_mode" value="yes"/> in the testcase works as expected. Please correct this.`

bjrr:
@brettleehari selenium 2 regression is attached but not selenium 3. Since this feature combines Selenium 3 support as well due to the merge conflict as stated: "We identified and changed the logic so that both the functionality is available to the user. Headless mode and selenium 3.0 support." 

Regression for Selenium 3 is missing. Unable to approve.

brettleehari:
@bjrr, 
Unable to upgrade the system I use for regression to selenium 3.0. Shall try alternate ways to share the regression for selenium 3.0.

kaheichan:
Attached failed regression log in Jira
Regression failure is the same one as reported in WAR-1548
Can't put this in release 3.5 as not able to verify compatibility with Selenium 3.0 support

To @brettleehari 
For future testing, I'd suggest using virtualenv to create a virtualenv to install selenium 3.0 and run the regression in it
the setup for virtualenv shouldn't take more than 2 mins

If virtualenv isn't install, can use pip install virtualenv
If sudo access is not given, can use pip install -U virtualenv
which will install virtualenv under current user

sanikakulkarni:
Pull Request has been approved by @navvrnd which was later dismissed on account of new commits.

kaheichan:
Due to major change to the code to support Selenium 3.0
A newer version of headless mode support has been implemented in #334

PR #52, opened 2017-06-28
Feature/war 1260: katana: Add bypassed checks in testdata files back into the testdata controller

keenan35i:
added basic checks.


keenan35i:
contains Bugfix/war 1259: katana: fix Iter Type value in cli data file

PR #55, opened 2017-06-29
WAR-1157: Generate JUnit file after each step

venkat1077:
Changes made:
1. Save the type of the file being executed by Warrior as a separate key, value pair in Warrior data_repo
2. After executing each step, create new junit file with updated details and replace existing(if any).
3. Fetch the values of 'resultsdir & logsdir' from testcase level property tag in junit file


venkat1077:
> Hi @venkat1077
> Looks like one of the testsuite in pj_retry isn't able to generate the html file correctly, can you check if the junit for that is correct?

@kaheichan made changes to fix it, results in WAR-1157 jira ticket.

kaheichan:
Ran pj_rest and confirm that step result generate before testcase ends

PR #56, opened 2017-06-29
Feature/WAR-746 Exectype with multiple rules and complex expression

kaheichan:
Change to Exectype element in step
Now support If/If Not exectype with:
1. Multiple rules, default status calculation will use logical AND with all rules
2. Each rule can use equal, not_equal to compare all supported datatype
in addition to that, greater_equal, greater, smaller_equal, smaller can be used to compare int or float datatype
3. Each rule can have its own else action and it will get triggered immediately if the rule fails
4. An optional expression at exectype attribute level is supported, user can use expression to build complex logic like:
1 or 2 and 3
1 and 2 | (3 and 4)
1 and (2 or (2 & 3))
the number stands for the index of rule inside exectype element
& | and or are supported, note that & symbol is not supported in xml, it has to be escaped with &amp;
Katana is handling that now so only manual user needs to escape the xml
5. An optional else attribute is supported at exectype level
this will get triggered if the decision is not execute and no else action has been triggered at rule level

Testing:
please run wftests/warrior_tests/suites/framework_tests/cond_var/ts_exectype_rules_expression.xml
it has covered simple exectype, multiple rules and multiple rules with expression scenario


warrior/WarriorCore/Classes/kw_driver_class.py
```diff
@@ -201,7 +201,6 @@ def get_argument_as_keywords(self):
     def execute_method_for_keyword(self):
         """Executes a method corresponding to keyword """
 
-        Utils.config_Utils.set_datarepository(self.data_repository)
```
satyakrish: why is this line being removed?

kaheichan: The same line is added in testcase_driver in L40
It gives reference to data_repo between tc_driver and kw_driver_class
and since we can't call step from outside, tc will be the smallest file warrior can execute from

satyakrish: in that case can you remove line 226 under the method execute_function_for_keyword as well

kaheichan: This is addressed

venkat1077:
opted out as Arvind from offshore is reviewing it.

bjrr:
Katana is being updated, WAR-912 is being developed so merging this.

kaheichan:
Rerun the test and verified it pass

warrior/WarriorCore/exec_type_driver.py
```diff
@@ -11,65 +11,349 @@
...
+    elif operator == "gt":
+        return True if get_object_from_datarepository(exec_condition) > exec_cond_var else False
+    elif operator == "le":
+        return True if get_object_from_datarepository(exec_condition) <= exec_cond_var else False
+    elif operator == "lt":
+        return True if get_object_from_datarepository(exec_condition) < exec_cond_var else False
+
+    pNote("Unknown error occur when deciding value, please check condition value of the step",
+          "Error")
+    return False
```
navvrnd: If there are any changes in the the name of get_object_from_datarepository or the parameter names, we have to make changes in multiple places. Instead the following code would be more efficient in my opinion:
do_operator = {'ge': lambda x,y: x>=y,
                          'gt': lambda x,y: x>y,
                          'le': lambda x,y: x<=y,
                          'lt': lambda x,y: x<y,
                          }
return do_operator[operator](get_object_from_datarepository(exec_condition), exec_cond_var)

warrior/WarriorCore/exec_type_driver.py
```diff
@@ -11,65 +11,349 @@
...
+        result = True if get_object_from_datarepository(exec_condition) == exec_cond_var else False
+    elif status and operator == "ne":
+        result = True if get_object_from_datarepository(exec_condition) != exec_cond_var else False
+    elif status and operator in ["ge", "gt", "le", "lt"]:
+        result = math_decision(exec_condition, exec_cond_var, operator)
+    else:
+        pNote("Execution condition failed for expected value: {} , operator: {}, actual value: {}"\
+              .format(exec_cond_var, operator,
+                      get_object_from_datarepository(exec_condition)), "WARNING")
+
```
navvrnd: 'eq' and 'ne' can also be handled in math_decision instead of duplicating it here.

warrior/WarriorCore/exec_type_driver.py
```diff
@@ -11,65 +11,349 @@
...
+    exec_condition = rule.get("Condition", None)
+    exec_cond_var = rule.get("Condvalue", None)
+    else_action = rule.get("Else", None)
+    if else_action is not None and else_action.upper() == "GOTO":
+        else_action = rule.get("Elsevalue")
+
+    # Check for math operator
+    support_operators = ["ge", "gt", "le",
+                         "lt", "eq", "ne"]
+    operator = rule.get("Operator", "eq")
```
navvrnd: can we have a global var with this list instead of duplicating this list in most of the functions here? I think the dictionary that I mentioned in math_decision can be taken globally and then used in different functions appropriately.

warrior/WarriorCore/exec_type_driver.py
navvrnd: verify_data in data_Utils.py also uses similar logic which can be shared as a common function and re-used instead duplicating in multiple places.

warrior/WarriorCore/exec_type_driver.py
kaheichan: please refer to L72, eq and ne can be used in other data type and math_decision is only used when data type is int/float
hence the separation

warrior/WarriorCore/exec_type_driver.py
kaheichan: I will put this in #245

warrior/WarriorCore/exec_type_driver.py
kaheichan: I will put this in #245

kaheichan: Change has been made in
https://github.com/warriorframework/warriorframework/pull/245

warrior/WarriorCore/exec_type_driver.py
kaheichan: Change has been made in
https://github.com/warriorframework/warriorframework/pull/245

PR #58, opened 2017-06-30
WAR-864 : Keyword Sequencing Tool Phase 1

priyaarun23:
Keyword Sequencing Tool Phase 1 : 
GUI : WAR-882
- Separate screen for Keyword Sequencing Tool.
- XML Object creation with User Inputs.
- Check for overwritten of Wrapper Keyword check.
- Tooltip for all required fields.
- Copy Subkeyword option like Testcase.
- Mandatory Fields handled.
- New Table layout with Insert , Edit and Delete options.

Python : WAR-864
- Parsing XML object from GUI.
- Posting up the values in Keyword Sequencing Template.
- wrapper kw would be created by grouping the keywords which is available in our warrior framework 
- Created wrapper kw would be listed while creating test cases by selecting the driver which had been already provided by the user while creating the wrapper kw.


priyaarun23:
@bjrr @brettleehari @sanikakulkarni 

Discussed with @kaheichan , for the Travis failure of pylint from katana.py which is not applicable as it checks for print_utils(Supported only in Warrior).

Hence requesting to review the same.

priyaarun23:
Travis has been passed. 
Kindly review the same. This ticket is at our second priority list.

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+@route('/searchkw', method='POST')
+def searchkw():
+    """
+    This method returns the ActionFile Path of the selected Driver.
+    """
+    value = parseString("".join(request.body))
+    indented_xml = "".join(value.toprettyxml(newl='\n'))
+    corrected_xml = remove_extra_newlines_char_xml(indented_xml)
+    corrected_xml = corrected_xml
+    with open('obj.txt', 'w') as f:
```
kaheichan: relative paths won't work if user isn't launching katana at the current working directory

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+        f.write(corrected_xml)
+    tree = ET.parse('obj.txt')
+    root = tree.getroot()
+    driver_name = root.text
+    global gpysrcdir
```
kaheichan: Please avoid using global variable

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+    class_list_new = []
+    import_action_list = []
+    sample = []
+    py_files = ""
+    doc_string_value = []
+    kw_list_1 = []
+    xmlobj = parseString("".join(request.body))
+    indented_xml = "".join(xmlobj.toprettyxml(newl='\n'))
+    corrected_xml = remove_extra_newlines_char_xml(indented_xml)
+    with open('output.txt', 'w') as f:
```
kaheichan: Same issue, relative path won't work

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+    method_names = []
+    arg_name_list = ""
+    object_list_new = []
+    class_list_new = []
+    import_action_list = []
+    sample = []
+    py_files = ""
+    doc_string_value = []
+    kw_list_1 = []
+    xmlobj = parseString("".join(request.body))
```
kaheichan: This part of code is similar to the one in the above function, please wrap them in the same helper function

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+    """
+    This method returns the ActionFile Path of the selected Driver.
+    """
+    value = parseString("".join(request.body))
+    indented_xml = "".join(value.toprettyxml(newl='\n'))
+    corrected_xml = remove_extra_newlines_char_xml(indented_xml)
+    corrected_xml = corrected_xml
+    with open('obj.txt', 'w') as f:
+        f.write(corrected_xml)
+    tree = ET.parse('obj.txt')
```
kaheichan: What's the purpose of parsing the xml, write to a tmp file and re-read it again?
Is is possible to parse it once with the correct format? Or have the formatting done in memory and not in file.

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+                                          'WrapperName')
+
+    ActionFile = getChildTextbyParentTag('output.txt', 'Details', 'ActionFile')
+    Description = getChildTextbyParentTag('output.txt', 'Details',
+                                          'Description')
+
+    # To check if the user provided wrapper keyword name already exists in action file
+    ActionFile = ActionFile.strip()
+    with open(ActionFile, 'r') as files:
+        for line in files:
```
kaheichan: files.readlines() can replace the for loop and the split under

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+    doc_string_value = []
+    kw_list_1 = []
+    xmlobj = parseString("".join(request.body))
+    indented_xml = "".join(xmlobj.toprettyxml(newl='\n'))
+    corrected_xml = remove_extra_newlines_char_xml(indented_xml)
+    with open('output.txt', 'w') as f:
+        f.write(corrected_xml)
+    tree = ET.parse('output.txt')
+    root = tree.getroot()
+    WrapperName = getChildTextbyParentTag('output.txt',
```
kaheichan: I can't see the reason why we need an output.txt here act as a temporary xml that stores action information, can it be a dict/other datatype that store the same info in memory?

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+        for line in files:
+            line = line.split()
+            if "def" in line:
+                next_word = line[line.index("def")+1].split("(")[0]
+                method_names.append(next_word)
+
+    if WrapperName in method_names:
+        checkval = "yes"
+    else:
+        checkval = "no"
```
kaheichan: Please either document this block of code or split it into different functions

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+    value = False
+    tree = ET.parse(datafile)
+    root = tree.getroot()
+    node = root.find(pnode)
+    if node is not None:
+        child = node.find(cnode)
+        if child is not None:
+            value = child.text
+            return value
+        else:
```
kaheichan: no need for else here, can just return value at the end

katana/katana.py
```diff
@@ -90,6 +93,207 @@ def readconfig():
...
+            for _, val in enumerate(object_list_new):
+                val = val.decode('utf-8')
+                f2.write('\n' + "        " + val)
+            for line in io.open('kw_seq_temp', 'r'):
+                f2.write('\n' + "    " + line)
+        f2.close()
+    return checkval
+
+
+def getChildTextbyParentTag(datafile, pnode, cnode):
```
kaheichan: To avoid excessive fileIO, please modify this helper function to take in xml etree instead of filename

priyaarun23:
@kaheichan We have developed this tool with the code from the new layout of katana. And the side bar icon works and fits fine over there. PFA for the same.
icon 1
icon2

katana/katana.py
kaheichan: @Annam16 
I tried launching katana using
python katana/katana.py
When I tried to save wrapper keyword, it gives me
Traceback (most recent call last):
  File "/home/ka/Desktop/warriorframework/katana/bottle.py", line 861, in _handle
    return route.call(**args)
  File "/home/ka/Desktop/warriorframework/katana/bottle.py", line 1741, in wrapper
    rv = callback(*a, **ka)
  File "katana/katana.py", line 252, in parsexmlobj
    for line in io.open('kw_seq_template', 'r'):
IOError: [Errno 2] No such file or directory: 'kw_seq_template'
127.0.0.1 - - [22/Aug/2017 10:00:43] "POST /parsexmlobj HTTP/1.1" 500 853
{
  "response": "no"
}
127.0.0.1 - - [22/Aug/2017 10:00:45] "GET /checkfileexistwithsubdir/test.xml/testcase/none HTTP/1.1" 200 22

kaheichan: Also regarding this comment:
If I launch katana outside of katana directory, it will try to create the out.txt file under my cwd, which has a chance of overwriting existed file

Annam16: I understand the issue @kaheichan . I will address it. Thank you.

katana/katana.py
Annam16: Hi @kaheichan 
The purpose for these lines(149 to 154) are to check the user provided wrapper kw name is already exists in user provided action file while saving the wrapper kw.  And here we are searching for the word "def" in each line by splitting it with split(). But the purpose of files.readlines() is reads in the whole file at once and splits it by line(split('\n')).  So I have used "split()" instead of "files.readlines()". Referred this link https://www.peterbe.com/plog/blogitem-040312-1 for understanding it.

kaheichan: in such case
To do it in pure text file way
please use if line.strip().startswith("def ") to check if it is actually a def before splitting it into a list of string
or use if line[0] == "def"
and then extract the method name

Or as I mentioned before
Please refer to
https://docs.python.org/2/library/ast.html
https://greentreesnakes.readthedocs.io/en/latest/nodes.html#function-and-class-definitions
to find function def directly

Annam16: Sure @kaheichan . As you said,  I have used line.strip().startswith("def ") to do the operation. Thank you.

katana/katana.py
Annam16: addressed

katana/katana.py
Annam16: done

katana/katana.py
Annam16: removed this method since I have not required it to get the values.

katana/katana.py
Annam16: removed this method since I have not required it to get the values.

katana/katana.py
Annam16: Done

katana/katana.py
Annam16: Done. created a new method "get_correct_xml_and_root_element" and used it

katana/katana.py
Annam16: Done

katana/katana.py
Annam16: Done. Added comments for each block

katana/katana.py
Annam16: Working on this comment.

Annam16: Done @kaheichan

katana/katana.py
Annam16: Addressed @kaheichan

Annam16:
@kaheichan I have addressed all the review comments. Kindly review.

bjrr:
@brettleehari I am delegating the review responsibility to you. Please review /ask your team members to review/test and once you approve and are satisfied that this pull request is good to go, please notify me to merge.

katana/katana.py
```diff
@@ -91,6 +93,208 @@ def readconfig():
...
+    """ To get the correct xml format from Katana UI and to get it's root element """
+    indented_xml = "".join(value.toprettyxml(newl='\n'))
+    corrected_xml = remove_extra_newlines_char_xml(indented_xml)
+    tree = xml.etree.ElementTree.fromstring(corrected_xml)
+    return tree
+
+
+@route('/parsexmlobj', method='POST')
+def parsexmlobj():
+    """
```
venkat1077: Break down this function into smaller units. Create separate methods to perform repeated operations(ex.  line 202 to 215 and line 223 to 234 looks similar). Use more meaning variable names ex. kw_name instead of kw_list(line 158).

katana/katana.py
```diff
@@ -91,6 +93,208 @@ def readconfig():
...
+                                arg_value = argument.text
+                            arg_list = argument_name + "=" + "\'" + arg_value + "\',"
+                            arg_name_list += arg_list
+                    kw_list_1.append("{0}.{1}({2})".format(class_list, kw_list, arg_name_list))
+                    doc_string_value.append("The keyword {0} in Driver {1} has a defined arguments "
+                                            "{2} You must want to send other values through data "
+                                            "file".format(kw_list, driver, arg_name_list))
+                    arg_name_list = ""
+        sum_val = 1
+        for _, new in enumerate(doc_string_value):# To have the doc string in the order with
```
venkat1077: use of enumerate here??

katana/katana.py
```diff
@@ -91,6 +93,208 @@ def readconfig():
...
+                            if arg_value is None or arg_value is False:
+                                arg_value = argument.text
+                            arg_name = argument_name + "=" + "\"" + arg_value + "\""
+                            kw_list_1.append("{0}.{1}({2})".format(class_list, kw_list, arg_name))
+                            # Forming doc string for wrapper keyword by using sub keyword details
+                            doc_string_value.append("The keyword {0} in Driver {1} has a defined "
+                                                    "arguments {2} You must want to send other "
+                                                    "values through data file".format(kw_list,
+                                                                                      driver,
+                                                                                      arg_name))
```
venkat1077: This string is used in multiple places. Define this in the beginning of the function and use it when required with dynamic substitution.

priyaarun23:
@venkat1077 
For your review comment - "Detailed message in Katana after creating wrapper kw(ex. where we can find the newly created wrapper keywords) would be helpful for the users" ::: Would be taken up in next phase of Sequencing Tool as this was not been a part of Phase 1 requirement.

priyaarun23:
@sugan19 
"In wrapper keyword details the tooltip of driver name should give more info that, this is the file in which the wrapper keyword will be saved" is almost similar to Venkat's comment. This would be taken up in next phase of Sequencing Tool as this was not been a part of Phase 1 requirement.

katana/katana.py
kaheichan: @priyaarun23 are the code related comments addressed? Would like to know if I should proceed to review again

brettleehari:
@bjrr ,
I am not the reviewer for the ticket, I am assigned for this ticket. Kindly do not add me as a reviewer.

bjrr:
@brettleehari I adding you as reviewer because I have delegated that responsibility to you. You will have to approve this before its is merge. My understanding is that this pull request was complete, passed internal review since it was sent to Rich team review. What is holding the approval?

priyaarun23:
@venkat1077 : Addressed all your review comments.

katana/katana.py
```diff
@@ -90,6 +93,227 @@ def readconfig():
...
+    katana_py_file = katana_exe.split('/')
+    if len(katana_py_file) > 1:
+        katana_py_file = katana_py_file[:-1]
+        katana_py_file = '/'.join(katana_py_file)
+        template_dir = os.getcwd() + '/' + katana_py_file + '/kw_seq_template'
+        template_dir_1 = os.getcwd() + '/' + katana_py_file + '/kw_seq_temp'
+    else:
+        template_dir = 'kw_seq_template'
+        template_dir_1 = 'kw_seq_temp'
+    return template_dir, template_dir_1
```
venkat1077: use (i) os.sep instead of "/", (ii) os.path.join to form path, (iii) os.path.dirname to get the dir name from the filepath

katana/katana.py
```diff
@@ -90,6 +93,227 @@ def readconfig():
...
+
+    if WrapperName in method_names:
+        # To check if the user provided wrapper keyword name already
+        # exists in action file
+        checkval = "yes"
+    else:
+        checkval = "no"
+        Subkeyword = tree.find('Subkeyword')
+        subkw_list = Subkeyword.findall('Skw')
+        doc_string_value, subkw_list, object_list_new, import_action_list, kw_name_1 = check_action_file_name(subkw_list, ActionFile)
```
venkat1077: this line exceeds 100 chars

katana/katana.py
```diff
@@ -90,6 +93,227 @@ def readconfig():
...
+    else:
+        template_dir = 'kw_seq_template'
+        template_dir_1 = 'kw_seq_temp'
+    return template_dir, template_dir_1
+
+
+def get_method_name(line):
+    """ To check whether the user provided wrapper keyword name already exists in user provided
+        action file from keyword sequencing tool UI screen"""
+    if line.strip().startswith("def "):
```
venkat1077: this condition will pass if a docstring line starts with 'def ' which is not correct.

katana/katana.py
```diff
@@ -90,6 +93,227 @@ def readconfig():
...
+                    line = line.replace('$wdesc', 'Description not provided by the user')
+                line = line.replace('$kw_name_1', str(kw_name_1))
+                f2.write(line)
+        final_obj_list = set(object_list_new)
+        final_import_list = set(import_action_list)
+        with open(ActionFile, 'a+') as f2:
+            for _, vals in enumerate(final_import_list):
+                vals = vals.decode('utf-8')
+                f2.write('\n' + "        " + vals)
+            for _, val in enumerate(final_obj_list):
```
venkat1077: enumerate not required in line 179 & 182

katana/katana.py
```diff
@@ -90,6 +93,227 @@ def readconfig():
...
 
+@route('/searchkw', method='POST')
+def searchkw():
+    """
+    This method returns the ActionFile Path of the selected Driver.
+    """
+    value = parseString("".join(request.body))
+    tree = get_correct_xml_and_root_element(value)
+    driver_name = tree.text
+    driver_file = gpysrcdir + '/ProductDrivers/' + driver_name + ".py"
```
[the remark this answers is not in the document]
navvrnd: done

katana/katana.py
navvrnd: done

katana/katana.py
navvrnd: done

katana/katana.py
navvrnd: done

katana/katana.py
navvrnd: Done, please check.

navvrnd: Changed the logic and removed the repeated operations. Please check.

katana/config/config.json
```diff
@@ -1,11 +1,11 @@
...
+  "projdir": "", 
+  "xmldir": "", 
+  "idfdir": "", 
+  "pythonpath": "", 
+  "testdata": "", 
+  "testsuitedir": "", 
   "engineer": ""
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
navvrnd: reverted to the original files.

navvrnd:
@venkat1077 addressed the old comments and 1 and 2 part of the new comments. The third point is fixed in war-1579 hence it works fine now. For the 4th part, I feel it is not necessary because of the following reason:
not importing the whole utils and checking if the module already exists and import if necessary is sheer overhead which is not worth the returns as Python already does optimizations if the same module is imported more than once.
 >>> %timeit import os
 1000000 loops, best of 3: 446 ns per loop
 >>> %timeit if 'os' in sys.modules: import os
 1000000 loops, best of 3: 547 ns per loop

Please check and approve

katana/keyword_sequencer_template
```diff
@@ -0,0 +1,44 @@
...
+            action_methods = inspect.getmembers(action_class, inspect.isroutine)
+            func = [item[1] for item in action_methods if item[0] == keyword][0]
+            output = func(action_class_obj, **kw_args)
+            if isinstance(output, tuple):
+                update_datarepository(output[1])
+                curstatus = output[0]
+            else:
+                curstatus = output
+            pNote(keyword_result_msg.format(keyword, curstatus))
+            if str(curstatus).upper() in ['ERROR', 'EXCEPTION']:
```
venkat1077: ERROR and EXCEPTION are two different statuses, do not change it.

navvrnd: Each individual keyword can have status of Error or Exception, but the final result of the wrapper keyword would be Error. Will document this.

venkat1077: Use case: I have couple of keywords in my wrapper and one(or both) of its result is 'EXCEPTION', as a user i would expect the overall result to be 'EXCEPTION'. If we get mix of ERROR and EXCEPTION statuses, overall status can be ERROR and that should be documented.

brettleehari:
Hi @bjrr ,
               This ticket is ready to be merged. Kindly do the needful.

bjrr:
Merging per Hari's request. (Venkat is on PTO )

PR #65, opened 2017-07-10
WAR-1293 : Fields Overwritten in Suite

priyaarun23:
Resolved the Engineer & Last Updated Fields overwritten, when we open an existing suite.


priyaarun23:
@sanikakulkarni 

When the user opens an existing(already saved) Suite file, there must be two scenarios :
	1.To view the details of the saved XML file : In this case user wants to read the saved details of the XML which has the engineer field of who have "created the file" and the last updated field which has the "time of last updation/creation".
	
	2.To update an existing XML file : In this case when user click on save button , the Engineer & Last Updated fields could be saved with the current system Date Time and the Engineer name from the configuration page. This happens on the controller part.
	
Say for example , you have created the sample.xml with the engineer field as "Sanika" & Last Updated field as the "time of your last updation/creation". When I open this sample.xml after 5 days it should show the Engineer as your name as you have created the sample.xml file and the Last Updated field as the time of the file creation/updation.Instead if that renders my name and my current system time , then I may mislead about the details of the saved file.

NOTE : Discussed with user community over here & they are in need of the old details when they view the saved files.

Kindly correct me if I am wrong.

sanikakulkarni:
Hi @priyaarun23,

Let me discuss this with the team and get back to you.

priyaarun23:
@sanikakulkarni 

I would like to know for any update on the same.

priyaarun23:
@sanikakulkarni 
Awaiting further updates.

priyaarun23:
As commented by Sanika, this fix would be taken up after a cohesive design is agreed upon during the (re)designing of Testcase app.

Hence closing this pull request.

And also re-opening the corresponding ticket and making it as related to WAR-1326 and WAR-1214.

PR #66, opened 2017-07-11
WAR-1141: selenium:not able to launch the browser with url if it is not the part of world wide web

brettleehari:
Current behaviour : For url that are not part of "WWW" as computed by urllib2 package of python, we stop navigating to that particular url an mark the step as failed.
Change : urllib2 result has no impact of the step and on whether we should navigate to that particular url. Instead we warn the user based on urllib2 responses.


Annam16:
Hi @sanikakulkarni  .
I have tested this fix with the test case which is available in warrior_fnc_tests(tc_browser_launch_url_check.xml) and it is failing . it requires some changes.Attached the log in JIRA.

warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py
```diff
@@ -178,27 +179,21 @@ def go_forward(self, browser_instance=None):
...
+                print_warning("The Status code for url : {} is {}".format(url, get_status_code))
         except urllib2.HTTPError as http_error:
             print_error("URLError: {} reason: ({}) status code: {}".format(url, http_error.reason, http_error.code))
-            status = False
         except urllib2.URLError as url_err:
             print_error("URLError: {} reason: ({})".format(url, url_err.reason))
-            status = False
-        if status == False:
-            print_error("Incorrect URL provided")
-        return status, url
```
sanikakulkarni: Hi @brettleehari,
Changing the number of arguments being returned is not advisable as it breaks backward compatibility for anyone making use of this function outside of the WSelenium keywords.

warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -182,10 +182,8 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
                     output_dict[browser_fullname] = browser_inst
                     if "url" in browser_details and browser_details["url"]\
                             is not None:
-                        result, url = self.browser_object.check_url(browser_details["url"])
-                        if result == True:
-                            result = self.browser_object.go_to(url,
-                                                               browser_inst)
+                        self.browser_object.check_url(browser_details["url"])
```
sanikakulkarni: There is not point calling the function check_url() in line 185 as it always is returning None.

sanikakulkarni:
Reply to: 

```Hi @sanikakulkarni .
I have tested this fix with the test case which is available in warrior_fnc_tests(tc_browser_launch_url_check.xml) and it is failing . it requires some changes.Attached the log in JIRA.```

Hi @Annam16,

Please let me know why this pull request was approved if tc_browser_launch_url_check.xml is failing. Please let me know if tickets have been created to address any issue found through/by the code changes made to satisfy WAR-1141

File: warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py (hunk 1 above)
Comment (brettleehari) [inline warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py]: Agreed and changed.

**Comment by Annam16** [comment]
Hi @sanikakulkarni 
I have approved this pull request even though "tc_browser_launch_url_check.xml" failed because, it has to be a separate pull request in bitbucket in warrior_fnc_tests. I have added this as a comment and approved it. 

No new tickets has been raised related to war-1141 fix. To make "tc_browser_launch_url_check.xml" pass/valid I have added this tc in WAR-1303 itself.

And this test case alone is enough to test this fix, since it contains non-www based url.

File: warrior/Actions/SeleniumActions/browser_actions.py (hunk 2 above)
Comment (brettleehari) [inline warrior/Actions/SeleniumActions/browser_actions.py]: The bug is url navigation has nothing to do check_url(). So check_url should not impact warrior attempting to navigate to the specified url. 
Check url just checks the function and no navigation is performed in this function.
Check_url is based on urllib2 module, which can validate public urls with http/https as schemes. 

The current purpose of check_url is to send warning/info/error to the user with appropriate message to the user.

**Comment by brettleehari** [comment]
@sanikakulkarni 
I have added functional tests for the fix with the same ticket and i did not create a new ticket for editing a functional test. 
Moreover since we are reverting back a feature, all we have to do it flip the status of the functional tests which were created for the earlier feature. Hope this clarifies everything. Kindly review and let me know if you have any more comments.

File: warrior/Actions/SeleniumActions/browser_actions.py [hunk 3]
```diff
@@ -184,12 +186,16 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
...
                     browser_fullname = "{0}_{1}".format(system_name, browser_details["browser_name"])
                     output_dict[browser_fullname] = browser_inst
-                    if "url" in browser_details and browser_details["url"]\
-                            is not None:
-                        result, url = self.browser_object.check_url(browser_details["url"])
-                        if result == True:
-                            result = self.browser_object.go_to(url,
-                                                               browser_inst)
+                    url = browser_details["url"]
+                    if url is not None:
```
sanikakulkarni: Hi @brettleehari,

I want to list down usecases of what should fail and what should not:

1. www.google.com: Pass
2. https://www.google.com: Pass
3. file:///path/to/file.html: Pass if file exists
4. file:///path/to/file.html: Fail if file does not exist
5. httpz://www.google.com: Fail
6. http://wwww.google.com: Fail
7/ http://googsle.com: Fail

bjrr: @brettleehari  what is the status of the above comments?

brettleehari: We had a call and explained to Sanika that we cannot do the operation as mentioned in sheet with examples. 
Addressed all the concerns.

sanikakulkarni:
Hi @brettleehari,

I am posting an email from a conversation that happened between us about this ticket:

1.	The check_url function needs to be removed from usage as it “opens” the url to retrieve the status. This implementation is faulty and can result in failed testcases (in cases where the URL should not be opened more than once). The check url function can exist in our utils (backward compatibility), but a disclaimer should be added in the check_url() docstrings that explains clearly what this functions does.
2.	As far as navigating to a url (that is incorrect and would result in subsequent step failures), the browser_launch and the navigate_to_url keyword should not ‘care’ about the navigation results and instead a brand new keyword (navigate_to_url_with_verification) should be added that lets the user give a url to it and then also provide a way to test if the url navigation went through (by giving a tag name or page title etc) to verify the page with. Now this keyword would then be able to navigate to the given url and then verify if the navigation was successful.
3.	Proper documentation should be added into the docstrings for browser_launch and navigate_to_url keywords to let the user know that these keywords would only navigate to urls and not verify if the navigation was successful and that if they want such verification, then they should use the navigate_to_url_with_verification keyword

This is to maintain consistency between review comments - and to keep track of the resolution reached about this PR's design.

- Sanika

warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py
```diff
@@ -177,27 +178,29 @@ def go_forward(self, browser_instance=None):
             self.current_browser.forward()
 
     def check_url(self, url):
-        """To check whether the user provided url is valid or not."""
+        """To check whether the user provided url is valid or not.
+           Returns:
+                    1. status(bool)= True / False.(Whether the url can be reached)
+                    2. url : The actual url itself
```
sanikakulkarni: Hi @brettleehari,

Please add documentation here that clearly states states that this function would internally "open" a URL to validate it.

Thanks,
Sanika

sanikakulkarni:
Hi @brettleehari,

According to point (2), we should be adding a keyword - navigate_to_url_with_verification - to let user perform verification after going to a URL.

Has a new ticket been created for this?

Thanks.

sanikakulkarni:
Hi @brettleehari,

Since it has been decided that browser_launch and navigate_to_url would not "care" about the validity of URL, we need clear documentation in those kws' docstrings which explicitly mention this.

This documentation has only been added for browser_launch. Can this also be added to navigate_to_url?

Thanks.

warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -47,7 +49,8 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
                        url=None, ip=None, remote=None, element_config_file=None,
                        element_tag=None):
         """
-        This will launch a browser.
+        The Keyword would launch a browser and Navigate to the url, if provided by the user.
+        The keyword does not validate the url provided by the user.
```
sanikakulkarni: Hi @brettleehari,

Can line 53 be emphasized? Like it can be made all caps or have an underline?

Thanks.

warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -182,21 +185,23 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
...
-                        if result == True:
-                            result = self.browser_object.go_to(url,
-                                                               browser_inst)
+                    url = browser_details["url"]
+                    if url is not None:
+                        urlschema = urlparse(url)
+                        if urlschema.scheme:
+                            result = self.browser_object.go_to(url, browser_inst)
+                        else:
+                            status = False
```
sanikakulkarni: Hi @brettleehari,

Instead of setting the status to False in line 197, the result should be set to False - eventually in line 206 the status will become False anyway because we are and-ing result and status there.

result refers to a single iteration the for loop (each browser opened) while status is the status of the entire kw.

Thanks.

warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py
brettleehari: Done! Explicitly added a note

warrior/Actions/SeleniumActions/browser_actions.py
brettleehari: @sanikakulkarni I have underlined as per your comment. Is this Warrior convention ? @bjrr What is your view ?

sanikakulkarni:
Selenium regression attached in JIRA ticket. Travis passing. Comments addressed. PR approved.

PR ready to be merged.

PR #70, opened 2017-07-12
Test rule checker with a copied actual pull request

kaheichan:
This is a test pull request, please review #69


kaheichan:
The actual rule checker branch went in #69

PR #72, opened 2017-07-14
WAR-1255: response stored in data repository through response_ref get stored with control characters

navvrnd:
response from cli commands stored through response_ref and pattern gets stored with control characters which is causing issues while retrieving later. This was tested with the local team facing the issue and was working fine. For testing, one can take a test case which will store data with control characters in data repository and then retrieve later without any issues.


kaheichan:
@navvrnd  Please provide title and description for this one, #73, #74, #75

Annam16:
@navvrnd can you add a detailed description in the JIRA ticket as well as how to test this fix in pull request description?

Annam16:
@navvrnd  Can you add sample test case to test this fix in JIRA

Annam16:
@navvrnd Can you please add the below details in JIRA ticket.
1. failed test case and it's related files.
2. Failure log before this fix.
3. Pass log after this fix. 
4. How you conclude that the test case  failed due to control characters because in ticket description it is mentioned that "not able to retrieve it in another test step"

warrior/Framework/Utils/cli_Utils.py
```diff
@@ -642,20 +642,23 @@ def _get_response_dict(details_dict, index, response, response_dict):
...
         if resp_pat_req is not None:
             # if the requested pattern not found return empty string
-            reobj=re.search(resp_pat_req, response)
-            response=reobj.group(0) if reobj is not None else ""
-            pNote("User has requested saving response. Response pattern required by user is : {0}".format(resp_pat_req))
-            pNote("Portion of response saved to the data repository with key: {0}, value: {1}".format(resp_ref, response))
+            reobj = re.search(resp_pat_req, response)
+            response = reobj.group(0) if reobj is not None else ""
+            # removing non-ascii/control characters (line feed) from response
+            response = "".join([c for c in response if 31 < ord(c) < 127])
```
venkat1077: Can you check if you can use "rem_nonprintable_ctrl_chars" method from testcase_utils_class file for removing ascii control chars.

navvrnd: The function "rem_nonprintable_ctrl_chars" only removes a subset of  control characters. I wanted to make it generic. May be we can replace that with this code, but I don't what might break. Please do comment.

venkat1077: 'rem_nonprintable_ctrl_chars' method removes chars with hex values other than 'x20-x7E, x09 & x0A' which is equivalent to accepting decimal values: 32-126,  9(horizontal tab) & 10(new line). I see the  difference is only 2 chars(horizontal tab & new line) which we should allow IMO.

warrior/Framework/Utils/data_Utils.py
```diff
@@ -750,7 +750,12 @@ def convert2type(value, data_type='str'):
     """
     type_funcs = {'str': str, 'int': int, 'float': float}
     convert = type_funcs[data_type]
-    return convert(value)
+    try:
+        cvalue = convert(value)
+    except Exception as exception:
+        print_exception(exception)
```
Annam16: 1. It would be better if we add the warning message instead of printing the whole exception.
2. I see couple of exceptions , one is we are raising and the other one is because of returning "None". Do necessary changes. 
3. Multiple return statements inside single method. we can assign it to a variable called "status" and handle that value in all the methods where this method is been called already.

navvrnd:
made the changes requested. Please review.

navvrnd:
cli full regression log attached in the jira ticket. The following tests failed in the full regression which does not look like issues due to this change:
1. connect: connect_call_connect_telnet_multi_subsys, connect_call_connect_telnet_custom_keystroke, connect_abort_as_error
2. cli_global_attr: tc_global_title_no_subsys, tc_global_title_one_subsys, tc_global_title_multi_subsys
3. cli_local_attr: tc_local_title_no_subsys
4. cli_other_attr: tc_cmd_retry_attrib
5. log_verify_on_suite: log_verify_on

navvrnd:
@kaheichan addressed comments in the jira ticket. Please check and approve.

kaheichan:
Hi @navvrnd 
More comment has been added in the Jira ticket
the TODO is:
1. Have a better error message when substitution went wrong
2. Add a testcase to test this error

kaheichan:
Upon further investigation in the dummy-td file and the attached td file, I found out that in the dummy-td file
user specifies a regular expression resp_pat_req="[\d]+\s" to capture the response from server.
\s in the resp_pat_req captures the non-printable character “\n” in the response “2017041011\n”
When the user uses the response as an integer conditional value, Python tries to cast “2017041011\n” to an int, which fails because \n cannot be converted into integer.

This is a user input error that we cannot prevent.
What we can do is when we are trying to use the saved value as conditional value and converting it into integer, we should remove the non-printable character before the conversion.

The current fix in WAR-1255 will cause issues in the future when the user actually wants to capture non-printable character such as newline.
Since the fix removes all non-printable characters, a scenario where user wants to capture a multi-line response will be distorted because we remove the new-line character in this fix.
Please update the fix and let us know immediately so we can review and merge asap.

navvrnd:
Hi @kaheichan addressed all the comments. Please check and approve.

warrior/Framework/Utils/data_Utils.py
```diff
@@ -1446,8 +1458,15 @@ def subst_var_patterns_by_prefix(raw_value, start_pattern="${",
                                     str(raw_value[k]).replace(
                                         search_obj.group(), 'None'))
                     except SyntaxError:
-                        print_error("Syntax Error - " +
-                                    error_msg2.format(string, value))
+                        from WarriorCore.Classes.testcase_utils_class import TestcaseUtils
```
kaheichan: Please put import statement at the beginning of the file

warrior/Framework/Utils/data_Utils.py
```diff
@@ -1446,8 +1458,15 @@ def subst_var_patterns_by_prefix(raw_value, start_pattern="${",
...
+                        tuc_obj = TestcaseUtils()
+                        print_info("Checking after removing the illegal characters")
```
kaheichan: Please provide a more detailed message
"Cannot convert value: {} into correct format, removing non-printable characters, will attempt conversion again"

warrior/Framework/Utils/data_Utils.py
navvrnd: done.

warrior/Framework/Utils/data_Utils.py
kaheichan: Hi @navvrnd 
The code looks good, I run the new testcase and verified it works
however cli_proj is failing and it shouldn't
please refer to the internal PR for more info

kaheichan:
Hi @navvrnd 
The code looks good, I run the new testcase and verified it works
however cli_proj is failing and it shouldn't
please refer to the internal PR for more info

navvrnd:
Hi @kaheichan I have attached all pass full regression log in the ticket after commenting out list_range test case which was causing hang. Please let me know, what else to be done for this ticket.

navvrnd:
@kaheichan reran and attached in the ticket. Looks like the code for WAR-1439 is still not yet in which is causing one failure. Please let me know, if any thing else needs to be done from my side.

kaheichan:
Saw the change and verified the fix, thank you @navvrnd

PR #74, opened 2017-07-14
Feature/war 991: Implement wtag feature at framework level

navvrnd:
Please check the design doc from the jira ticket. Please note that file option would not work here as expected, as it is implemented separately in war-1172.


Annam16:
@navvrnd Please resolve merge conflicts

warrior/Actions/CommonActions/common_actions.py
```diff
@@ -228,4 +237,48 @@ def set_env_var(self, var_key=None, var_value=None, filepath=None,
 
         return status
 
+    def get_values_from_datafile(self, system_name, strvar, langs, states,
+                                 currencys, ramspace, configfile, intvar,
+                                 file_config):
+        """get the values from datafile
```
Annam16: @navvrnd Add description for each argument in documentation string

wftests/warrior_tests/testcases/framework_tests/tc_wtag_unit_tests.xml
```diff
@@ -0,0 +1,80 @@
...
+<Testcase>
+	<Details>
+		<Name>tc_wtag_unit_tests</Name>
+		<Title>test wtag feature</Title>
+		<Engineer>Arvind Kumar Sethuraman</Engineer>
+		<Date>2017-06-09</Date>
+		<Time>15:51</Time>
+		<State/>
+		<InputDataFile>../../data/data_wtag.xml</InputDataFile>
+		<!-- <InputDataFile>../../../warrior_fnc_tests/warrior_tests/data/framework_tests/data_wtag.xml</InputDataFile> -->
```
Annam16: please remove this commented line since it contains our local repository name.

warrior/Actions/CommonActions/common_actions.py
```diff
@@ -228,4 +237,48 @@ def set_env_var(self, var_key=None, var_value=None, filepath=None,
...
+        """
+        def check_type(var, datatype):
+            """check that vars are of correct datatype
+            """
+            vartype = type(var)
+            if vartype is not datatype:
+                print_error('is not a {}, but {}'.format(datatype, vartype))
```
Annam16: Can you please add the error message more clear?

warrior/Actions/CommonActions/common_actions.py
navvrnd: done.

wftests/warrior_tests/testcases/framework_tests/tc_wtag_unit_tests.xml
navvrnd: done.

warrior/Actions/CommonActions/common_actions.py
navvrnd: done.

navvrnd:
addressed all review comments, please review and approve.

warrior/Actions/CommonActions/common_actions.py
```diff
@@ -228,4 +237,59 @@ def set_env_var(self, var_key=None, var_value=None, filepath=None,
 
         return status
 
+    def get_values_from_datafile(self, system_name, strvar, langs, states,
+                                 currencys, ramspace, configfile, intvar,
+                                 file_config):
```
venkat1077: I see this kw is for testing purpose not for generic use, name 'get_values_from_datafile' says it is to get values but it actually checks the type and returns the status and also the argument names are not generic & the docstring doesn't have any info of how the datafile should be written(using type) for this kw. What would happen if the user doesn't give 'type' tag in the datafile and use this keyword with wtag.

wftests/warrior_tests/data/data_wtag.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" ?>
```
venkat1077: Move this data_file to framework_tests directory since the tc is placed inside framework_tests.

warrior/WarriorCore/Classes/kw_driver_class.py
```diff
@@ -157,33 +158,65 @@ def get_mandatory_arguments(self):
             args.remove('self')
         return args
 
+    def get_credential_value(self, arg, system):
+        """get the value of arg in test data file corresponding to system
```
venkat1077: It is data_file, right?(typo)

warrior/WarriorCore/Classes/kw_driver_class.py
```diff
@@ -157,33 +158,65 @@ def get_mandatory_arguments(self):
...
+        """
+        datafile = self.data_repository['wt_datafile']
+        var = arg
+        if not hasattr(self, 'tag_dict'):
+            self.tag_dict = data_Utils.get_credentials(datafile, system)
+        if isinstance(arg, basestring) and arg.startswith("wtag"):
+            var = arg.split("=")[1]
```
venkat1077: strip whit spaces when splitting arg value.

warrior/WarriorCore/Classes/kw_driver_class.py
venkat1077: Doc string says that this method is to get the value from the 'test data file' but suppose to be 'data file'.

Annam16:
@navvrnd 
1. Add the repo details like link where the repo is available(wft_kw_repo)
2. Add the pull request link for sample test cases which is been added in bitbucket
3. This feature has to be added in regression. If any new functionality break this "wtag" feature we would not be knowing it .
4. Not able to identify which tag is failed . The printed error message is not conveying it exactly.
5. when I ran the given test case which is in bitbucket with a normal data file without "type tag", it just thrown an error message only for the tag "langs" which contains the value of list. For the remaining tags also it should throw the same error. It would be better if we throw that as a warning message instead of error message.

navvrnd:
@Annam16 
Addressed points 1 and 2 in the description. For point 3, we don't have a proper keyword which can test the wtag feature and hence have created a separate keyword in wft_kw_repo which tests all the wtag scenarios. Hence this can not be put in normal regression suites. Fixed point 5 in wft_kw_repo repo. For point 4, as I have mentioned earlier, it is better to integrate with war-1172 (file type option) once it is available in the repo.

warrior/Actions/CommonActions/common_actions.py
venkat1077: Done

venkat1077: Completed.

wftests/warrior_tests/data/data_wtag.xml
venkat1077: Completed.

warrior/WarriorCore/Classes/kw_driver_class.py
venkat1077: Completed.

warrior/WarriorCore/Classes/kw_driver_class.py
venkat1077: Completed.

warrior/WarriorCore/Classes/kw_driver_class.py
```diff
@@ -157,33 +158,66 @@ def get_mandatory_arguments(self):
...
+            var = arg.split("=")[1].strip()
+        if var in self.tag_dict:
+            value = self.tag_dict[var]
+            # substitute environment/datarepo variables in the value and return
+            if isinstance(value, (basestring, list, dict)):
+                return data_Utils.substitute_var_patterns(value)
+            else:
+                return value
+        else:
+            return None
```
kaheichan: If var is not found in tag_dict, it should print an error message
please add one here unless there is an error message print somewhere else associate with this

navvrnd: It is handled in get_values_for_(mandatory|optional)_args...

warrior/Actions/CiRegressionActions/ci_regression_actions.py
```diff
@@ -329,3 +329,99 @@ def delete_tmp_dir(self):
         """
         path = data_Utils.get_object_from_datarepository("parallel_exec_tmp_dir")
         return file_Utils.delFolder(path)
+
+    def check_kw_arg_type(self, str_value, int_value, float_value, bool_value,
```
kaheichan: If check_type in check_values_from_datafile can achieve the same test coverage, then we don't need 2 functions

warrior/Actions/CiRegressionActions/ci_regression_actions.py
```diff
@@ -329,3 +329,99 @@ def delete_tmp_dir(self):
...
+            status = False
+        if type(dict_value) is not dict:
+            print_error(err_msg.format(dict_value, "dict", type(dict_value)))
+            status = False
+        if type(file_value) is not file:
+            print_error(err_msg.format(file_value, "file", type(file_value)))
+            status = False
+        return status
+
+    def check_values_from_datafile(self, system_name, strvar, langs, states,
```
kaheichan: Please enhance the functionality to actually compare the value obtained with an expected value

warrior/Actions/CiRegressionActions/ci_regression_actions.py
navvrnd: check_type is for arguments starting with type_ (like int_, file_ etc.,) whereas the other is wtag units. I have moved the changes from war-1172 to here, so that there is less confusion.

warrior/Actions/CiRegressionActions/ci_regression_actions.py
navvrnd: done. Please check.

warrior/Actions/CiRegressionActions/ci_regression_actions.py
kaheichan: I see, please update the documentation for this keyword to clarify:
This keyword is intended to test the type prefix for keyword arguments
when an argument name has a type_ prefix, the variable type will become the type specified in the type_ prefix

kaheichan: Also the keyword name is confusing to me
it'd be clearer if it says check_kw_arg_type_prefix

warrior/Actions/CiRegressionActions/ci_regression_actions.py
```diff
@@ -329,3 +329,99 @@ def delete_tmp_dir(self):
...
+                                   currencys, ramspace, configfile, intvar,
+                                   file_config):
+        """get the values from datafile
```
kaheichan: This keyword description should be
Verify the datatype of the value read from the datafile using either the tag or wtag feature

warrior/Actions/CiRegressionActions/ci_regression_actions.py
```diff
@@ -329,3 +329,99 @@ def delete_tmp_dir(self):
...
+            vartype = type(var)
+            status = True
+            if vartype is not datatype:
+                print_error('{} is expected to be {} type, but found to be of '
+                            '{} type'.format(varname, datatype, vartype))
+                status = False
+            return status
+        status = True
+        datafile = Utils.config_Utils.datafile
+        tc_filepath = os.path.dirname(data_Utils.get_object_from_datarepository(
```
kaheichan: Please add inline comments to indicate which part of the code is doing the check for a particular type of value
for example
`#this block checks if value is string type
status = check_type(strvar, "strvar", str) and status
#this block checks if value is list type
status = check_type(langs, "langs", list) and status
`

navvrnd: done.  Please check and approve.

warrior/Actions/CiRegressionActions/ci_regression_actions.py
navvrnd: done.

warrior/Actions/CiRegressionActions/ci_regression_actions.py
kaheichan: Hi @navvrnd 
What about L431-448
I am still not sure what that part of the code is trying to test

warrior/Actions/CiRegressionActions/ci_regression_actions.py
kaheichan: @navvrnd 
Looks like this comment get hidden but not addressed, can you update the description please?

warrior/Actions/CiRegressionActions/ci_regression_actions.py
navvrnd: Hi @kaheichan that is to test the existing tag functionality which should not be broken because of the new wtag functionality

warrior/Actions/CiRegressionActions/ci_regression_actions.py
```diff
@@ -329,3 +329,125 @@ def delete_tmp_dir(self):
...
+                # this block checks if the contents of file type variable is expected
+                print_error("contents of the file {} is <<{}>> which does not match expected"
+                            " <<{}>>".format(file_value, actual_contents, file_contents))
+                status = False
+        return status
+
+    def check_values_from_datafile(self, system_name, strvar, langs, states,
+                                   currencys, ramspace, configfile, intvar,
+                                   anotherfile):
+        """get the values from datafile
```
kaheichan: This comment get hidden...copying to make a new one
This keyword description should be
Verify the datatype of the value read from the datafile using either the tag or wtag feature

warrior/Actions/CiRegressionActions/ci_regression_actions.py
kaheichan: I see, understood.
A suggestion for the valid file check, let's add another line of os.path.isfile to enhance the check

navvrnd:
Hi @kaheichan please let me know, if any thing else is pending from my side on this ticket.

navvrnd:
Hi @kaheichan Why can't this be pushed for this release? Please let me know, if there is any thing pending from my side.

navvrnd:
Hi @kaheichan I have attached the latest logs in the ticket after merging the latest changes. Please review and approve.

warrior/Framework/Utils/data_Utils.py
```diff
@@ -160,7 +161,17 @@ def get_credentials(datafile, system_name, myInfo=[], tag_name="system",
         output_dict = {}
         if len(myInfo) == 0:
             for child in element:
```
satyakrish: per war-1165  the datype support should be implemented for both get_credentials and getSystemData .. I dont see it implemented for getSystemData

satyakrish: data type support in the below code addition is implemented only if myInfo is empty it should be implemented for te case when users requests specific tags by providing specific values in myInfo.

warrior/WarriorCore/Classes/kw_driver_class.py
```diff
@@ -158,33 +159,66 @@ def get_mandatory_arguments(self):
...
```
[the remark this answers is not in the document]
navvrnd: 1 - added inline comments explaining the need. Basically, this is the place where we are converting wtag var to the actual value if wtag var is used
2 - this is for the case where the mandatory var is not menioned in the args but it need to be picked up from wtags as per our wtags design doc.
3 - yes, it makes it easier for me to access it. Please let me know, if you have any other suggestions
  a. do you want to move the initialization part to the constructor? But this requires accessing system_name arg and calling get_credentials function. 
  b. from point a, if the initialization is moved to the constructor part, then the destruction can be done in destructor part. Basically, need a mechanism to clean up tag_dict after processing both mandatory and optional args. Doing it in get_values_for_mandatory_args would make it unavailable in get_values_for_optional_args function.
   c. yes, but it needs to be passed around in atleast these three functions - get_values_for_mandatory_args, get_values_for_optional_args and get_credential_value

4 - I am checking if the arg is starting with wtag and then only splitting on "=". What advantage would I get if I split on "wtag="? Atleast with the other approach, I can handle white spaces around the operators which won't be possible in this approach.

warrior/Framework/Utils/data_Utils.py
navvrnd: Sure Sathya, will work on this. Can you please provide any cases with non-empty myInfo for my testing?

warrior/WarriorCore/Classes/argument_datatype_class.py
```diff
@@ -10,25 +10,43 @@
...
 
-import ast
-import traceback
-from Framework.Utils.print_Utils import print_error, print_info
 
 class ArgumentDatatype(object):
     """This is the class that gets the data type of an argument
-    supplied by the user and stores it in the data repository """
+    supplied by the user and stores it in the data repository.
+    The file option here is implemented separately in war-1172,
```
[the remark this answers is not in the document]
navvrnd: done, removed it.

warrior/Framework/Utils/data_Utils.py
satyakrish: @navvrnd 
for non empty myInfo you can refer to Actions/DemoActions/demo_actions.py testset_calibration keyword line 142-143 
            credentials = get_credentials(self.datafile, call_system_name, \
                                ['calibration', 'user', 'location', 'testdata'])

wftests/warrior_tests/testcases/framework_tests/tc_kw_arg_type_prefix.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" ?>
```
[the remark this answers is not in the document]
navvrnd: done

wftests/warrior_tests/testcases/framework_tests/tc_wtag_unit_tests.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" ?>
```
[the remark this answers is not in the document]
navvrnd: done

warrior/Framework/Utils/data_Utils.py
navvrnd: done. Added test case for checking myInfo and getSystemData from the demo actions as well

warrior/WarriorCore/Classes/kw_driver_class.py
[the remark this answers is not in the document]
navvrnd: @satyakrish
#1: 
 - wtag  feature would not work if system name is not specified
 - if it is not specified in first place the wtag feature can not be used for the arguments before the system name argument is specified
 - if code is added to check for system name in the args list and then process each argument, then this adds unnecessary cycles for each keyword
#2: 
targ = arg_kv[args] if args in arg_kv else args
if args is arg_int and the user has passed 'wtag=intvalue' for this arg then targ would be assigned 'wtag=intvalue' otherwise it would be arg_int and if this arg_int is defined in tag_dict, that should be picked up as per our design document, which is what this get_credential_value() does. 
#3: This is indeed advisable practice. Please refer https://stackoverflow.com/questions/45486395/what-happens-when-you-initialise-instance-variables-outside-of-init."As a dynamic language, Python behaves in exactly the same way wherever you set an attribute, and __init__ is not in any way privileged when it comes to creating attributes."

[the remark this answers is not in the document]
navvrnd: Can we discuss this in a call? Please let me know a convenient time when we can discuss this.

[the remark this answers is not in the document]
satyakrish: @navvrnd  yes pls schedule for a call from outlook during my working hours.

navvrnd:
@satyakrish made the changes as per your comments. Please review and approve. Logs attached in the ticket.

satyakrish:
@navvrnd
cc: @bjrr

The changes look good, but while reviewing it this time I observed another issue. With the currently implementation do we assume that a keyword would always have mandatory arguments? If so we should make changes to implement the case where the keyword has all optional arguments as well. Because there is no check to stop a developing a keyword which has all optional arguments including the system name. Can you verify this case and make changes accordingly.
Also please update the test cases to include such a scenario.

navvrnd:
@satyakrish cc: @bjrr Made the requested changes. Please check and approve.

satyakrish:
@navvrnd : the changes you made are not sufficient? The purpose of the change was to make sure that if a keyword has all optional arguments , wtag substitution should work fine? but you have not covered it at all...

You have updated your keyword to have all optional arguments.
1- but wtag substitution is not taking place anywhere.  Please make sure to use wtag in this case and make sure it works fine as well... you should use wtag substitutions as default values of an argument and also as a value provided by user in the test case.

2- you should also test the case where
- all arguements are optional (lets say there are 5 arguements)
- system name should be the third argument (in both keyword signature and the arguements provided in the testcase)
- make sure at least one argument preceding system name has wtag substitution  (lets say the first argument) and also one argument after system_name should have wtag substitution (lets say 5 argument) and both cases should work fine.

navvrnd:
@satyakrish Made the required changes and updated the ticket with the cli full regression, wtag and kw prefix logs. Please check and approve.

PR #75, opened 2017-07-14
Bugfix/war 1299: input data file provided from cli is not taken when run in mockrun

navvrnd:
attached the full cli regression run log in the jira ticket.


navvrnd:
added full regression run log after merge changes to the jira ticket.

PR #95, opened 2017-07-27
Bugfix/war 575 : REsolving ENV variable when user uses tag=value in testcase args. 

brettleehari:
1. The fix is for resolving ENV variable in data file which has a unique tag, which was defined in testcases with the feature tag=tag(in data file) 
2. The fix will resolve ENV variable not only for REST but for all modules.
3. After ENV we now have REPO support in Warrior. Added REPO resolution support for tag=tag(datafile) feature.

Added tests to cover:
est-1 details :
Step-1: I set the env variable url
Step-2: I dont provide url a value in Testcase instead defined url=ENV.url in data file. 
Result : It substituted appropriately.
Test -2 details :
Step-1: I set the env variable url
Step-2: I provide url a value in Testcase itself, but the value is a ENV value. Defined url=ENV.url in testcase. 
Result : It substituted appropriately.
Test -3 details :
Step-1: I set the env variable url
Step-2: I provide url a value "tag=url1" in Testcase, defined url1=ENV.url in datafile. 
Result : It substituted appropriately.

CLI regression results available in WAR ticket. 
Rest tests in WFT runs successfully.


warrior/Framework/Utils/data_Utils.py
```diff
@@ -1298,10 +1300,11 @@ def get_user_specified_tag_values_in_tc(datafile, system_name, **kwargs):
     for element in kwargs:
         in_list.append(element)
     credentials = get_credentials(datafile, system_name, in_list)
-
     for element in kwargs:
         if kwargs[element] is not None:
             credentials[element] = resolve_argument_value_to_get_tag_value(datafile, system_name, kwargs[element])
+    credentials = sub_from_env_var(credentials)
```
satyakrish: The correct place to handle env, data repo substitutions would be in the function resolve_argument_value_to_get_tag_value, because the current function get_user_specified_tag_values_in_tc internally uses resolve_argument_value_to_get_tag_value.

If we don't handle this support in resolve_argument_value_to_get_tag_value then users who directly use this function will not get this functionality.

I have added some comments to resolve_argument_value_to_get_tag_value  as well please incorporate them also.

warrior/Framework/Utils/data_Utils.py
```diff
@@ -1266,13 +1266,15 @@ def resolve_argument_value_to_get_tag_value(datafile, system_name,
     if element_value_in_argument.startswith("tag"):
```
satyakrish: the check should be .startswith("tag=")

warrior/Framework/Utils/data_Utils.py
```diff
@@ -1266,13 +1266,15 @@ def resolve_argument_value_to_get_tag_value(datafile, system_name,
...
         tag_name = evaluate_tc_argument_value(element_value_in_argument)
         if tag_name is not False:
```
satyakrish: The else condition corresponding to this if should have a warning message that .. could nto find tag with the provided name = <name provide> hence returning the value as such.. or something like this

warrior/Framework/Utils/data_Utils.py
```diff
@@ -1266,13 +1266,15 @@ def resolve_argument_value_to_get_tag_value(datafile, system_name,
...
-            system_name_list = xml_Utils.get_matching_firstlevel_children_from_root(datafile, "system")
+            system_name_list = xml_Utils.get_matching_firstlevel_children_from_root(datafile, 
+                                                                                    "system")
             if system_name_list == [] or system_name_list is None or system_name_list is False:
```
satyakrish: please see if it good to add a print that system with provided name could not be found hence will return the value without tag= substituions

wftests/warrior_tests/testcases/rest_functional_tests/rest_variable_in_json_test.xml
```diff
@@ -1,233 +1,326 @@
...
-			<Description>This step tests the POST capability of REST</Description>
-			<Execute ExecType="Yes"/>
-			<context>positive</context>
-			<impact>impact</impact>
-		</step>
-	</Steps>
-
-</Testcase>
\ No newline at end of file
+        <Details>
```
sanikakulkarni: Changing an existing testcase in our established regression tests is not recommended. Please create a brand new testcase providing coverage to the functionality added in your pull request and add that to the suite.

warrior/Framework/Utils/data_Utils.py
brettleehari: I shall do it.

PR #96, opened 2017-07-27
Feature/WAR-1336 transform pylint.sh into python based script

kaheichan:


kaheichan:
This build shows how the report will look like

PR #98, opened 2017-07-28
WAR-1348 : Modularise runmode across Warrior. As a user, I want same tags for runmode wherever I use runmode functions such as RMT, RUP & RUF

brettleehari:
When user uses RMT in suite level, the option is ATTEMPTS. 
But when users do not use the tag MAX_ATTEMPTS which is used for RUP&RUF SUITES fail to run. 
REmoved the logic based on MAX_ATTEMPTS and simplified the code. 

Added a test with SUITE level RMT.


warrior/WarriorCore/testsuite_driver.py
```diff
@@ -335,14 +335,7 @@ def execute_testsuite(testsuite_filepath, data_repository, from_project,
...
     elif execution_type.upper() == 'RUN_MULTIPLE':
-        Max_Attempts = Utils.xml_Utils.getChildAttributebyParentTag(testsuite_filepath, 'Details', 'type', 'Max_Attempts')
-        Number_Attempts = Utils.xml_Utils.getChildAttributebyParentTag(testsuite_filepath, 'Details', 'type', 'Number_Attempts')
-        
-        if Max_Attempts == "":
-            execution_value = Number_Attempts
-        else:
-            execution_value = Max_Attempts
-        
+        execution_value = Utils.xml_Utils.getChildAttributebyParentTag(testsuite_filepath, 'Details', 'type', 'Number_Attempts')
```
satyakrish: please add a check here to see if execution value is a valid integer otherwise throw a warning and default it to 1 so that the test runs only once

brettleehari: Hi @satyakrish,
                          I agree to the above improvement comment that we have to evaluate whether user has provided numeric value to handle user error. It is an improvement from our current behavior where we trust that user provide whole numbers. 
I feel above suggestion is not a clean solution, currently there is bug and this ticket just fixes that bug. 
While working on the issue found few anomalies Kindly consider below points and if you still feel you want me to check value for RMT is integer i shall do it:

1. Just as you suggested change for RMT in SUITE global level, RUP & RUF requires changes too. 
2. We support Runmode in SUITE step level and we would require similar changes here too.
3. We support Runmode in TESTCASE GLOBAL and STEP level and requires same changes. 
4. Although we support RMT, RUP, RUF at 4 different levels(as mentioned above) their functionalities are the same but we use maxattempts, Number_attempts xml tags in SUITE, value tag in Testcases.

Given the above 4 points, I propose two changes:
1. RUNMODE functionalities shall carry the same tag across Warrior since the functionalities are exactly same. Slowly we can deprecate older tags or have both.  
2. Instead of handling user error such as validating whether a value is a whole number. How about using Django form fields to restrict users to only fill numeric else throw error. Because as a user I do not want to find out my user errors while executing but while developing.

satyakrish: @brettleehari 

- #1: we should go with both for backward comatibility, i dont think we can get rid of one at this moment.

- Agree that the error handling should be applied for all the cases like rup, ruf, rmt. 
Where should we do it? data validation  should be done at the front end for sure.. but we should not totally ignore error handling at warrior level. Because even today there are lot of users who use warrior directly without using the UI. 
I would suggest to raise a ticket for error handling at UI, and handle errors in warrior using this ticket.
If we encounter an un supported value then we should raise a warning that prints out what was expected and what was received and that default value will be used instead with the default value printed.

brettleehari:
Hello reviewers,
Work in progress, Please do not review for now. I shall let you know.

warrior/WarriorCore/common_execution_utils.py
```diff
@@ -12,8 +12,76 @@
...
+        1. filepath    = full path of the Testcase/suite/project xml file
+        2. step_tag = xml tag for group of step in the file
+        3. sub_step_tag = xml tag for each step in the file
 
+    """
+    step_list_with_rmt_retry = []
+    root = Utils.xml_Utils.getRoot(filepath)
+    steps = root.find(step_tag)
+    if steps is None:
+        print_warning("The file: '{}' has no steps "
```
kaheichan: can replaced the word steps with value of step_tag

warrior/WarriorCore/common_execution_utils.py
```diff
@@ -12,8 +12,76 @@
...
+                    for i in range(0, value):
+                        copy_step = copy.deepcopy(step)
+                        copy_step.find("runmode").set("value", go_next)
+                        copy_step.find("runmode").set("attempt", i + 1)
+                        step_list_with_rmt_retry.append(copy_step)
+                # only one step in step list, append new step
+                else:
+                    for i in range(0, value):
+                        copy_step = copy.deepcopy(step)
+                        copy_step.find("runmode").set("attempt", i + 1)
```
kaheichan: I know I wrote this more than a year ago...but I think this part also need to set the runmode value to go_next?

kaheichan: I think I wrote this more than 1 year ago...but even if this is the 1st step, it will need to set the value to next also

warrior/WarriorCore/common_execution_utils.py
```diff
@@ -12,8 +12,76 @@
...
+                        copy_step.find("retry").set("attempt", i + 1)
+                        step_list_with_rmt_retry.append(copy_step)
+                else:
+                    if runmode is not None:
+                        get_runmode = step.find('runmode')
+                        step.remove(get_runmode)
+                    for i in range(0, retry_value):
+                        copy_step = copy.deepcopy(step)
+                        copy_step.find("retry").set("attempt", i + 1)
+                        step_list_with_rmt_retry.append(copy_step)
```
kaheichan: line 48-81 contain 4 blocks of similar code
Since we are reworking the functionality
we should also wrap them in a common function/code block and reuse

warrior/WarriorCore/testsuite_driver.py
```diff
@@ -320,12 +264,59 @@ def execute_testsuite(testsuite_filepath, data_repository, from_project,
...
+                test_suite_status = sequential_testcase_driver.main(testcase_list, suite_repository,
+                                                                    data_repository, from_project,
+                                                                    auto_defects=auto_defects)
+                test_count = i * len(testcase_list)
+                testsuite_utils.pSuite_update_suite_tests(str(test_count))
+                if str(test_suite_status).upper() == "FALSE" or \
+                                str(test_suite_status).upper() == "ERROR":
+                    break
+
+    #The below ELIF is to preserve backward compatibility. The new logic is available/
```
kaheichan: If the value of execution_type hasn't changed, why would the runmode logic not handled by the original code?

warrior/WarriorCore/common_execution_utils.py
brettleehari: Done

warrior/WarriorCore/common_execution_utils.py
brettleehari: Done

warrior/WarriorCore/testsuite_driver.py
brettleehari: The Runmode is under execution type in the current design. So, I am not changing anything to the current logic to preserve backward compatibility. To bring uniformity across runmodes, I have introduced a tag similar to the tags available for runmode in suite_step, Project_step & Testcase_step in the Suite Global tags. Going forward users would see the new tags through Katana (This would be similar to all runmode options across Warrior). 
If this explanation doesn't help, I shall call you tomorrow to discuss.

warrior/WarriorCore/common_execution_utils.py
kaheichan: For step_list length > 1
there is a line
copy_step.find("runmode").set("value", go_next)
but it doesn't exist when step_list is empty

warrior/WarriorCore/common_execution_utils.py
kaheichan: For step_list length > 1
there is a line
copy_step.find("runmode").set("value", go_next)
but it doesn't exist when step_list is empty

brettleehari: Done

warrior/WarriorCore/common_execution_utils.py
kaheichan: Oops...What I mean is...step_tag is an xml element
and in the element there is a value specifying whether it is a step/testcase/testsuite
so the warning message can correctly say
The file X has no step/testcase/testsuite to be executed

warrior/WarriorCore/common_execution_utils.py
```diff
@@ -11,9 +11,92 @@
...
+    if step_tag is None:
+        print_warning("The file: '{}' has no steps to be executed".format(filepath))
+    else:
+        step_list = step_tag.findall(sub_step_tag)
+        # iterate all steps to get the runmode and retry details
+        for index, step in enumerate(step_list):
+            runmode, value = get_runmode_from_xmlfile(step)
+            retry_type, _, _, retry_value, _ = get_retry_from_xmlfile(step)
+            if runmode is not None and value > 0:
+                if len(step_list) > 1:
```
kaheichan: Looks like L70-79 are duplicated, any reason to keep both?

warrior/WarriorCore/common_execution_utils.py
```diff
@@ -11,9 +11,92 @@
...
+                    step_list_with_rmt_retry = append_step_list(step_list_with_rmt_retry, step,
+                                                                value, go_next, mode="runmode",
+                                                                tag="value")
+                else:
+                    go_next = len(step_list_with_rmt_retry) + value + 1
+                    step_list_with_rmt_retry = append_step_list(step_list_with_rmt_retry, step,
+                                                                value, go_next, mode="runmode",
+                                                                tag="value")
+            if retry_type is not None and value > 0:
+                if len(step_list) > 1:
```
kaheichan: Same here

kaheichan:
@brettleehari Looks like there is a merge conflict

kaheichan:
Ran the test and verified it work

bjrr:
@kaheichan Is there a Katana change needed for this ticket?

brettleehari:
@kaheichan Can you please share your reasons for rejecting the PR you approved. It would help me to resolve it at the earliest.

brettleehari:
Dear @kaheichan,
              Gentle reminder-2. This ticket is part of 3.5.0. Humbly Request to share comments so that I can work on this and deliver this item for the team.

kaheichan:
Hi @brettleehari 
A detailed comment has been made in the Jira ticket, please go ahead and check it out there

To summarize, please deliver bug fix and code improvements in 2 separate PR
Also if you decide to use this PR as the bugfix PR, please update the description of this PR as it is outdated

satyakrish:
transferred review to Ka on oct-6 hence removing myself.

sanikakulkarni:
Pull Request has been approved by @venkat1077  which was later dismissed on account of new commits.

kaheichan:
Github is preventing me from clearing the requested change...

ReshmaManohar:
Closing this PR. Will open a new PR for this issue.

PR #103, opened 2017-07-31
WAR-1340 : Links to open Testcase from Suite

priyaarun23:
Dynamically formed links of Testcase(from Suite) could be opened in Linux as well as Windows.

### DIFF
--- a/katana/static/app/controllers/projectCaptureCtrl.js
+++ b/katana/static/app/controllers/projectCaptureCtrl.js
@@ -11,8 +11,8 @@ limitations under the License.
 
 */
 
-app.controller('projectCapCtrl', ['$scope', '$http', '$routeParams', '$controller', '$location', '$timeout', 'fileFactory', 'saveasProjectFactory', 'getProjectFactory', 'setProjectFactory', 'subdirs',
-    function($scope, $http, $routeParams, $controller, $location, $timeout, fileFactory, saveasProjectFactory, getProjectFactory, setProjectFactory, subdirs) {
+app.controller('projectCapCtrl', ['$scope', '$http', '$routeParams', '$controller', '$location', '$timeout', 'fileFactory', 'saveasProjectFactory', 'getProjectFactory', 'setProjectFactory', 'getConfigFactory', 'subdirs',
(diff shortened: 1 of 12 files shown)


priyaarun23:
@sanikakulkarni 
Request you to review the same.

priyaarun23:
@sanikakulkarni Updated the fix according to the Design change.

**Old Design** :  Opening a case in suite opens up its corresponding XML in new browser tab.
**New Design** : Opening a case in suite opens up its corresponding Katana UI with the details in new browser tab.

Shown demo to the user and they agree for the fix.

Hence requesting you to review the same, as they are in need of this feature to ease their work.

**Note** : Fixed WAR-1350 in this ticket itself. _WAR-1350 : To open Suite from Project._

katana/static/app/controllers/projectCaptureCtrl.js
```diff
@@ -36,6 +36,36 @@ app.controller('projectCapCtrl', ['$scope', '$http', '$routeParams', '$controlle
...
+          url = filepath.split('..')[1];      
+          splitDir = url.split('/Suites')[1]; 
+          finalUrl = "#/testsuite"+splitDir+"/none";
+          window.open(finalUrl);
+
+       }
+
+        else{
+          
+          splitPath = filepath.split("/").pop(-1);
```
kaheichan: Need some comments for this block of code.

kaheichan:
verified functionality, pending review on code quality

priyaarun23:
Added Comment Lines.

kaheichan:
@priyaarun23 Saw the fix, it looks good
Just one last thing before approving, the error message for opening an non-existing tc is 500 undefined
Can we have a more meaningful error message for that? other than that the functionality of the PR is verified

priyaarun23:
@kaheichan : '500 undefined error' for the **non-existing file** is throughout the Katana application. This  could be handled in WAR-1222. Hence requesting you to approve this ticket.

kaheichan:
Hi @priyaarun23 
WAR-1449 is created in Jira for the error message

PR #118, opened 2017-08-07
Feature/war 851: adding support for installing dependencies through virtual environment

navvrnd:
Adding support for installing dependencies through virtual environment. One assumption is the user would have virtual env binary installed in his home dir at .local/bin/virtualenv. Otherwise the user has to provide the correct location in default_config file. Testing with virtual env option in default_config file and dependencies which are not normally allowed in the user environment but should pass in the virtual env environment.


navvrnd:
declining, since these are old changes bought forward by mistake.

PR #119, opened 2017-08-07
feature/WAR-851-github: adding support to add python dependencies in …

navvrnd:
…a virtual environment.
Testing: check warhorn with or without virtual env repo tag in default_config and with different dependencies for which the user does not have permission but can be installed through virtual env.
Providing different destinations in warriorframework repo tag should create the venv in those dir.


navvrnd:
pylint failed in travis since print is used in war_print_class, which is unavoidable since that is the only place we are using print. Hence request to please review and give comments.

warhorn/default_config.xml
```diff
@@ -16,6 +16,19 @@
         <dependency name="cloudshell-automation-api" install="yes" user="no"/>
     </warhorn>
 
+    <!-- Fill this out with the details of the virtual environment
+         to be used. Please note that if this is not used, then the warhorn
```
sanikakulkarni: Test.

sanikakulkarni: Done.

kaheichan:
Hi @navvrnd 
The code looks good, can you please update your branch with the latest develop (git merge or git rebase)
to get rid of the parallel exeuction test errors?

navvrnd:
Hi @kaheichan, merged the new changes. Please review and approve.

warhorn/warhorn.py
```diff
@@ -1103,6 +1070,21 @@ def get_base_path(node_name="warriorframework", **kwargs):
...
+    ve_name = get_attribute_value(node, 'name')
+    ve_loc = get_attribute_value(node, 'location')
+    ve_dest = destination + ve_name
+    print_info("ve_name: "+ve_name, logfile, print_log_name)
+    print_info("destination: "+ve_dest, logfile, print_log_name)
+    venv_cmd = os.path.expanduser(ve_loc)
+    subprocess.call([venv_cmd, "--system-site-packages", ve_dest])
+    venv_file = "{}/bin/activate_this.py".format(ve_dest)
+    execfile(venv_file, dict(__file__=venv_file))
+    return True
```
kaheichan: This function shouldn't only return True, there are multiple operations in here that can go wrong

warhorn/warhorn.py
```diff
@@ -1198,12 +1181,25 @@ def assemble_warrior():
 
     check_basic_requirements(logfile, config_file_name, console_log_name,
                              print_log_name, python_executable)
-    get_dependencies(logfile, print_log_name, config_file_name)
+    node = get_node(config_file_name, 'virtualenv')
+    if node is not False and get_attribute_value(node, 'activate') == 'yes':
+        if get_attribute_value(node, 'install') == 'yes':
+            install_depen('virtualenv==15.1.0', 'virtualenv', logfile, print_log_name)
```
kaheichan: There should be a change to avoid this kind of hardcoded value in code, not sure if it went in or not, but please try to avoid hardcoded version value

warhorn/warhorn.py
```diff
@@ -1198,12 +1181,25 @@ def assemble_warrior():
...
+        war_tag = get_node(config_file_name, 'warriorframework')
+        if war_tag is False:
+            print_error("warriorframework is a mandatory repo. Please add and rerun",
+                        logfile, print_log_name)
+            setDone(1)
+            getDone(logfile, print_log_name)
+        dest = get_attribute_value(war_tag, 'destination')
+        virtualenv_activated = activate_virtualenv(node, dest, logfile, print_log_name)
```
kaheichan: If activate is set to yes for virtualenv, I think it is reasonable to keep virtualenv activated after installation finish, that way user don't need to activate virualenv again before running warrior

warhorn/warhorn.py
navvrnd: execfile returns None if successful, otherwise raises exception, if any of the previous operations were a failure.

warhorn/warhorn.py
navvrnd: get_dict_with_versions from source/utils.py is loaded with hard coded versions for all the dependencies. Is there any plan to change this?

warhorn/warhorn.py
navvrnd: But, activating virtualenv is with in warhorn script, which would get cleaned up automatically, when you come out of the script.  We can keep the sub-process active if it can be back ground process, but here it can only be a foreground process and when the parent process completes, it would clean up the child processes as well along with the virtualenv environment. Can you elaborate on what you have in mind?

warhorn/warhorn.py
kaheichan: 1.  subprocess.call returns different exit code, please catch the non-success exit code and fail the status
2. As your comment mentioned, execfile will raise exception if fail, we need to catch the exception and safely return false

warhorn/warhorn.py
kaheichan: There is a plan to change that and I am not sure if that went in or not
@sanikakulkarni do you know more about this?

warhorn/warhorn.py
kaheichan: I am looking for the something similar to
https://stackoverflow.com/questions/6943208/activate-a-virtualenv-with-a-python-script
under
"""
However if you want to activate the virtualenv under the current python interpreter...
"""

navvrnd: I think, I have done the same thing in activate_virtualenv function as follows:
        venv_file = "{}/bin/activate_this.py".format(ve_dest)
        execfile(venv_file, dict(__file__=venv_file))
Am I missing something?

warhorn/warhorn.py
navvrnd: Ok, I have removed the version number to pick the latest. I guess, that should solve the problem.

warhorn/warhorn.py
navvrnd: ok, changed it.

warhorn/warhorn.py
kaheichan: I see, sorry for the confusion
What I am referring to is the terminal prompt change to (warrior_virtualenv) <prompt>
after the execution of warhorn
As I am not able to see it after executing warhorn
That's why I ask if it is possible to activate it right after the execution
Maybe not exiting the warhorn script will do the trick?

navvrnd: Yes, we can not achieve the ve prompt by exiting the warhorn script. Please approve or let me know, what you are expecting.

kaheichan: Hi @navvrnd 
I discussed with Brian about the activate attribute in the virtualenv tag
we agreed that the activate attribute isn't really necessary here if it only indicates whether to install dependencies inside the virtualenv or not
If the user gives a name for the virtualenv, then we should use the virtualenv
if the name field is blank, then we shouldn't use it
there's no need for an activate attribute
Can you please make the change for this?

navvrnd: Yes, done the changes. Please review and approve.

navvrnd: @kaheichan Can you please review this?

bjrr:
@brettleehari who from the offshore team is reviewing this? I need a review from your side.

warhorn/warhorn.py
```diff
@@ -1107,6 +1071,26 @@ def get_base_path(node_name="warriorframework", **kwargs):
...
+        venv_cmd = os.path.expanduser(ve_loc)
+        subprocess.check_call([venv_cmd, "--system-site-packages", ve_dest])
+        venv_file = "{}/bin/activate_this.py".format(ve_dest)
+        execfile(venv_file, dict(__file__=venv_file))
+        return True
+    except Exception as e:
+        print_error("Activating virtual env at {} resulted in exception {}".format(
+                                                ve_dest, e), logfile, print_log_name)
+        return False
+
```
venkat1077: Exit value should be set to 1 when activation fails.
use-case: Activation will fail when giving no or invalid value for 'location' attrib

brettleehari: Overall we need a result summary or result html for warhorn. This comment is out of scope of this review.

sanikakulkarni:
Hi @navvrnd,

Please let me know the status of this PR.

warhorn/warhorn.py
navvrnd: @venkat1077 done the changes. Please check and approve.

kaheichan:
Ran the latest code with virtualenv option enabled/disabled/commented
all works fine
also confirm when enabled, the dependencies are installed inside the virtualenv

PR #41, opened 2017-06-21
Feature/war-1108 Create HTML result file on the fly

keenan35i:
Contains:
- WAR-985 As a user, I want warrior version to be displayed in the HTML result file.
- WAR-987 As a user, I want to have filters in html result file 
- WAR-836 It would be better if the Test step number should also be displayed in the results .html

PR #120, opened 2017-08-07
Bugfix/WAR-1366 Multiprocessing hangs when queue is full

kaheichan:


kaheichan:
Issue has been found and it is caused by #41 in junit_class
attaching html object to junit object exceed its size and clog the multiprocessing queue
waiting for meeting to design solution

kaheichan:
Address in #124

PR #132, opened 2017-08-14
WAR-1356: Modify send_command(cli) utility to support both pexpect and WarriorConnect objects

venkat1077:
1. Move warrior_connect_class.py to WNetwork dir and rename it as warrior_cli_class.py
2. Create equivalent methods for all cli_Utils methods in warrior_cli_class
3. Now cli_Utils will only have wrappers to call the respective methods from warrior_cli_class

Tested by executing existing cli_regression.
CLI regression results & product team's test results are in jira ticket.


warrior/Framework/Utils/cli_Utils.py
```diff
@@ -174,125 +113,88 @@ def connect_telnet(ip, port="23", username="", password="",
...
-    if child.isalive():
-        if child.ignore_sighup:
-            child.ignore_sighup = False
-        child.close()
     return child
 
-@cmdprinter
-def send_command_and_get_response(sessionobj, prompt1, prompt2, command):
+
+def send_command_and_get_response(session_object, start_prompt, end_prompt, command):
```
satyakrish: do not change names of existing arguments.. because users may be using it as named arguments in their function calls like send_command_and_get_response(session_object, prompt1='.*', prompt2='\$', command=cmd) and changing name will break this function call.

warrior/Framework/Utils/cli_Utils.py
```diff
@@ -174,125 +113,88 @@ def connect_telnet(ip, port="23", username="", password="",
...
+    if isinstance(session_object, WNetwork.warrior_cli_class.WarriorCli):
+        _, response = session_object.send_command(start_prompt, end_prompt, command)
+    elif isinstance(session_object, pexpect.spawn):
+        wc_obj = WNetwork.warrior_cli_class.WarriorCli()
+        wc_obj.conn_obj = WNetwork.warrior_cli_class.PexpectConnect()
+        wc_obj.conn_obj.target_host = session_object
+        _, response = wc_obj.send_command(start_prompt, end_prompt, command)
+    else:
+        response = ""
+        print_warning("Unable to send the command since the session_object is not an "
```
satyakrish: change print statement as '....session object is not supported by warrior_cli_class...' instead of 'session_object is not an  instance of warrior_cli_class..'

PR #142, opened 2017-08-23
WAR-1133 : read requirements file to install dependencies

brettleehari:
Read the requirement file and add dependencies. 
Allow the user to provide their own requirement file, so that our requirement file is not edited.

Test : 1
Run warhorn.py after changing the values in requirement file and see if they are installed properly.
Run warhorn.py after deleting a key&value pair and see if the error is handled appropriately.
Provide another requirement file and run warhorn to see if warhorn ignores requirement.txt and takes the user defined file.
Provide a file which does not exist, see if warhorn falls back to default requirement.txt file. 
One can install all dependancy at one go through pip -r requirement.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,10 @@
+jira==1.0.3
+lxml==3.3.3
+ncclient==0.4.6
+paramiko==1.16.0
+pexpect==4.2
+pysnmp==4.3.2
+requests==2.9.1
+selenium==2.48.0
+xlrd==1.0.0
+cloudshell-automation-api==7.1.0.34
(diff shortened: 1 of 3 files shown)


brettleehari:
Hi @navvrnd ,
                             As discussed, I have two designs for the same ticket. The one you had pointed out is also available "https://github.com/warriorframework/warriorframework/pull/142".

bjrr:
@navvrnd Please complete review

brettleehari:
Tracked with alternate approach "https://github.com/warriorframework/warriorframework/pull/143"

PR #143, opened 2017-08-24
Another design for WAR-1133

brettleehari:
Developed two designs for the same tickets. 
The first one is available in : https://github.com/warriorframework/warriorframework/pull/142

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ requests==2.9.1
 selenium==2.48.0
 xlrd==1.0.0
 cloudshell-automation-api==7.1.0.34
+PyVirtualDisplay==0.2.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,10 @@
+jira==1.0.3
+lxml==3.3.3
+ncclient==0.4.6
+paramiko==1.16.0
+pexpect==4.2
+pysnmp==4.3.2
+requests==2.9.1
+selenium==2.48.0
+xlrd==1.0.0
+cloudshell-automation-api==7.1.0.34
--- a/warhorn/default_config.xml
+++ b/warhorn/default_config.xml
@@ -4,7 +4,7 @@
 
 <data>
     <warhorn name="Warhorn" dependency_filepath="Provide filepath, where version details for packages below are defined. By default the versions defined in requirements.txt at the base directory would be used">
-        <dependency name="jira" install="no" user="no"/>
+        <dependency name="jira" install="yes" user=""/>
         <dependency name="lxml" install="yes" user="no"/>
         <dependency name="ncclient" install="no" user="no"/>
         <dependency name="paramiko" install="yes" user="no"/>
@@ -14,6 +14,7 @@
         <dependency name="selenium" install="yes" user="no"/>
         <dependency name="xlrd" install="no" user="no"/>
         <dependency name="cloudshell-automation-api" install="yes" user="no"/>
+        <dependency name="PyVirtualDisplay" install="yes" user="no"/>
     </warhorn>
 
 	<warriorframework url="https://github.com/warriorframework/warriorframework.git"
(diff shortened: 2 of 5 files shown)


warhorn/source/utils.py
```diff
@@ -617,6 +602,22 @@ def install_depen(dependency, dependency_name, logfile, print_log_name,
...
+    filepath : The filepath from where version_package is derived.
+    """
+    version_package = False
+    if not os.path.exists(filepath):
+        filepath = "../requirements.txt"
+    with open(filepath, 'r') as requirements:
+        for line in requirements:
+            if package in line:
+                version_package = line
+    return version_package, os.path.abspath(filepath)
```
navvrnd: Please use this function to get only the version. May be create another function to get the actual file to be used.

warhorn/default_config.xml
```diff
@@ -3,8 +3,8 @@
 <!-- To read full documentation for this file, please refer to user_generated/config_sample.xml -->
 
 <data>
-    <warhorn name="Warhorn">
-        <dependency name="jira" install="no" user="no"/>
+    <warhorn name="Warhorn" dependency_filepath="Provide filepath, where version details for packages below are defined. By default the versions defined in requirements.txt at the base directory would be used">
```
bjrr: Hi @brettleehari,

The design for the dependency_filepath attribute was not mentioned anywhere in the design document.

warhorn/default_config.xml
```diff
@@ -3,8 +3,8 @@
...
+        <dependency name="jira" install="yes" user=""/>
```
bjrr: Hi @brettleehari,

Why was the "jira" dependency set to yes? We do not need the jira package any more - migrated to requests package many versions ago.

requirements.txt
```diff
@@ -0,0 +1,11 @@
+jira==1.0.3
```
bjrr: Hi @brettleehari,

The design document specifically mentions that the requirements.txt file should reside under warriorframework/warhorn and not directly under warriorframework. Keeping it under warriorframework directory implies that all the packages in the requirements.txt file are necessary for warrior to run - when that is not the case.

bjrr:
This code does not follow the design spelled out in the design document and email attached to the JIRA ticket by:

1. Lets user install "extra" dependencies
2. Additional "dependency_filepath" attribute
3. requirements.txt not inside the correct directory.

bjrr:
Hi @brettleehari,

The console logs for this feature were not attached to the Jira ticket.

bjrr:
Hi @brettleehari,

The design document also mentions creating testcases to test this feature. These testcases have not been added to the pull request.

warhorn/default_config.xml
brettleehari: It was a change while testing the ticket. Shall change

requirements.txt
brettleehari: @bjrr I am afraid that the design document clearly mentions what i have implemented. I do not agree with this comment.

brettleehari:
@bjrr Kindly open the pull request from closed state to open up for discussion.

warhorn/default_config.xml
brettleehari: I shall explain this over the call. We had discussed about this item.

requirements.txt
brettleehari: @bjrr Moreover, as per python pip requirement.txt conventions. The file is generally placed in the base directory. One possible change that we could do is to place it in warrior/ directory as these dependency apply for Warrior directory. Users who do not use warhorn assembler would only use requirements.txt having it in warhorn is not useful from user perspective. 

Main use of requirement.txt is pip funtionality of pip -r requirement.txt. I have designed it for warhorn and pip install -r <filename> (pythonic) usage. Both cases would work well :)

what do you think?

PR #151, opened 2017-08-30
Bugfix/war 1120: make tools options similar to warriorspace in warhorn

navvrnd:
Testing: check the different options for tools especially clean_install and clean_install_warrior should not work. And overwrite should overwrite in to warrior/Tools dir if already there otherwise leave as it is.


bjrr:
@brettleehari Who from the offshore team is reviewing this?

warhorn/warhorn.py
```diff
@@ -158,6 +158,8 @@ def overwrite_files(path, destination, overwrite, logfile, print_log_name):
 
     """
     sub_files, sub_folders = get_subfiles(path)
+    sub_files = filter(lambda f: not f.startswith('.'), sub_files)
```
kaheichan: What is the reason for this to be here? If it is just trying to avoid . and ..
it may miss some hidden folders/files that starts with . also

warhorn/warhorn.py
```diff
@@ -400,7 +379,16 @@ def clone_warrior_and_tools(base_path, current_dir, repo_root, **kwargs):
                                               console_log_name=console_log_name,
                                               print_log_name=print_log_name)
                 getDone()
-
+    overwrite = kwargs.get("overwrite")
+    dest_dir = os.path.join(base_path, "warrior/Tools/"+repo_name)
```
kaheichan: I think join functions accept more than 2 arguments, we can separate warrior Tools and repo_name into different parts

warhorn/warhorn.py
navvrnd: Is there a usecase for hidden files/folders to be copied? Here we are trying to avoid hidden files like git etc., from being copied.

masteret: Thank you for the reply, I was trying to figure why the code should be here, can you add an inline comment stating that these 2 lines are to avoid hidden files?
Right now we don't have any hidden files but we may need to be aware of these lines in the future if we introduce them

kaheichan: Thank you for the reply, I was trying to figure why the code should be here, can you add an inline comment stating that these 2 lines are to avoid hidden files?
Right now we don't have any hidden files but we may need to be aware of these lines in the future if we introduce them

navvrnd: Done. Please check and approve.

warhorn/warhorn.py
navvrnd: Done. Please check and approve.

kaheichan:
@navvrnd 
I am still having issue described in this comment
https://github.com/warriorframework/warriorframework/pull/151#pullrequestreview-66617419
When I try to clone the tools repo from internal repo it would fail
can you reproduce the issue and try to fix it?

brettleehari:
@kaheichan Able to reproduce the issue. 
Please correct me if i my below statement is wrong.
Like Warriorspace attributes in config file, Tools require attributes clone&overwrite to work exactly same as anyother clone or overwrite attribute in config file.
To add to it, I requested Arvind to add label attribute to Tools repo. 
Tools will now contain overwrite, clone & label attributes working as expected in line with Warriorspace or WARKEY repo.

navvrnd:
I have misunderstood the requirement, hence created a new pull request with the new understanding. Please review https://github.com/warriorframework/warriorframework/pull/240 instead of this. Closing this pull request.

PR #153, opened 2017-08-31
WAR-1311 : Way to insert Testcase in Suite

priyaarun23:
Testcase could be inserted while Creating/Editing a Suite.


priyaarun23:
Fix for WAR-1400 & WAR-1401 is also available in this ticket.

brettleehari:
image

Inserting a Test in Suite

PR #172, opened 2017-09-12
Katana dev ui kh

kamranHusain:
Removed small buggy items. No items changed in core area. Only pulled items from others plus my changes.


kamranHusain:
With latest changes for the demo

PR #176, opened 2017-09-15
WAR-1251 warrior user guide opens when warrior help is executed

sugan19:
Warrior user guide opens when ./Warrior --help command is executed.

current warriorframework path is fetched and opens the user guide if the terminal supports.


sugan19:
@bjrr @navvrnd 
Hi Reviewers, 

Few changes are required in the PR, will let you know once it is done.

PR #180, opened 2017-09-18
WAR-1252 Desktop notification is set after execuion completes

sugan19:
Desktop notification is set after the execution completes. It verifies the os_name and sends the notification command accordingly for the particular os.


sugan19:
@bjrr @satyakrish @venkat1077 
Hi Reviewers, 

Few changes are required in the PR, will let you know once it is done.

PR #190, opened 2017-09-21
WAR-954 : Auto-population of System & Subsystem 

priyaarun23:
From the user provided Input data file, System name would be populated in a dropdown.
Upon selecting the system name, the corresponding subsystems would be listed.


bjrr:
@brettleehari Should this be merged without further review from Richardson Team?

sanikakulkarni:
Hi @priyaarun23,

Please change to "if autopopulation 'is' needed
screenshot2
"

Thanks.

sanikakulkarni:
Hi @priyaarun23,

What was the design in the case when I open an existing testcae file containing an IDF path, but then click on "No Data File"?

Thanks.

sanikakulkarni:
Hi @priyaarun23,

Even if I check the "No data file" checkbox and open a new step,  am getting this alert:
screenshot1

Can this scenario be handled?

Thanks.

brettleehari:
@sanikakulkarni(Clarification:whether you want to rearrange error sentence ? or you meant otherwise)
Do you want the pop up error code to read 
"If Auto population is needed, Kindly provide the path and then Click on 'New step Button'" Instead of 
"Kindly provide the path and then Click on 'New step Button', If Auto population is needed, "

bjrr:
@priyaarun23 @brettleehari  Sanika's comment is simply stating that the 'is' is missing from the sentence:
Kindly provide the path and then click on 'New Step' Button if auto population **"is"** needed.

brettleehari:
@sanikakulkarni If you could see whether your change request are addressed. It would be great. If you are not satisfied with your change request then we will have monday to address your concerns.

sanikakulkarni:
Hi @priyaarun23,

Now, I can see that the data file path is clickable - that is, if I click on the data file path, that file is opened in another tab. (This is great, I did not know this existed) 

Now, what if I want to add a system that does not exist in the data file? To achieve this, I will have to open that datafile (presumably by clicking on it's path) add a system there, refresh this testcase page to see the system name pop up. This seems too complicated for the user. Can we devise a way in which this is simplified? May be we can let the user add a system that does not exist in the data in the dropdown for systems - and just add that system tag and name to the datafile (? this is up for discussion)

Please think about it from the perspective of the user

Thanks.

sanikakulkarni:
Hi @priyaarun23,

If I edit a step of an existing testcase (this has a value for the argument system_name), the dropdown for the system name remains empty - although, if that step is saved, the system_name value is retained. This was not happening before the latest commit. Please look into it.

Thanks.

priyaarun23:
Hi @bjrr @sanikakulkarni 
That doesn't seems to be a blocking issue. We could fix it separately. Hence I recommend to merge WAR-954.

Now in new commit, whenever the step is being edited with below scenario , the values of System/Subsytem will be cleared off for that particular step. If we edit step-1, other step values will not change as per the framework.
->Opening a saved TC with System/Subsystem values auto populated from a Input Data File.
->Edit a step and then check the "I don't want a data file" checkbox.
->System values would be reset and render as Textbox field for user input.
->Upon clicking on save that corresponding step will be saved with updated values. (As per the framework, editing Step-1 will update only that respective step and not all other steps.)

Thanks,
Priya (Oct 31)

sanikakulkarni:
Hi @priyaarun23,

Please checkout this comment: https://github.com/warriorframework/warriorframework/pull/190#issuecomment-340502074

priyaarun23:
Hi @sanikakulkarni ,

Addressed the review comment.
TC-> Edit Step -> Values of System & Subsystem will be populated for new as well as existing file.

Requesting you to review the same.

Thanks,
Priya A

sanikakulkarni:
Pull Request has been approved by @brettleehari which was later dismissed on account of new commits.

sanikakulkarni:
Hi @priyaarun23,

If I open a new file (it does not contain a path to the data file) and create a step (so the step is open in the step editor) - and realize that I can add a data file  to get the system list and therefore, I add that data file path. Now in this case the step that is already open in the step editor does not get updated to reflect the system name dropdown (i.e. it still shows a text box for system name); but if I open a new step, that step has the dropdown. Please look into this.

All other use cases work great.

Thanks,
Sanika

priyaarun23:
@sanikakulkarni ,

	To make user aware of the Data file values auto population, the popup (screenshot attached) has been provided when user clicks on 'New Step' button without entering the value in 'Input data file' (for new case).
	Upon reading the above said popup, user will be conscious about the system/subsystem auto-population functionality.
	
	Hence it is more explicit that below scenarios are not needed.
		1.New Testcase -> No data file -> Open New step -> Need a data file so adding it now.
		2.New Testcase -> No data file -> Open New step -> Need a data file so adding it now -> Feel 
                   like not needed so removing it -> goes on for infinite looping.
		3.New Testcase -> No data file -> Open New step -> Need a data file so adding it now -> 
                   values of system/subsystem name chosen from dropdown -> Clicking 'No data' now.
		
Few more not-so-needed scenarios can be added to above points.
The same has been updated in the design document as discussed with @brettleehari 

	NOTE : By considering all the scenarios and to make the design simple for the scope of ticket we have already mentioned in the design document that user should enter 'Input data file' path to get the system/subsystem auto values or the other way.

priyaarun23:
popup

sanikakulkarni:
Hi @priyaarun23,

Please document the above as known issues with this feature in the ticket. I will approve the PR as soon as this is done.

Thanks,
Sanika

priyaarun23:
@sanikakulkarni ,

Already added it in the design document.
Added the same in the ticket now.

Kindly do the needful.

priyaarun23:
@sanikakulkarni ,

Resolved the merge conflicts.
kindly do the needful.

sanikakulkarni:
Feature works as designed, hence approving.

PR #213, opened 2017-09-27
War-1439: regex for ip (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}) is mistaken with variable substitution in res_pat_req

sugan19:
In td file, the value for res_pat_req is given as a regex for an ip  (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}). If the user gives a vc_file in the testcase, then the '}' is mistaken as the end pattern for variable substitution and returns \d{1,3}\\.\\.\\. 

So modified the __expand_variable method in configuration_element_class.py
Attached the cli_regression in the ticket (telnet related TC failed in regression).
Attached sample file in the ticket.


warrior/Framework/ClassUtils/configuration_element_class.py
```diff
@@ -73,7 +76,8 @@ def __expand_variables(self, string):
 
             if match is None:
                 clean_partial += return_value[:return_value.find(check) + len(check)]
-                return_value = return_value.replace(return_value[:return_value.find(check) + len(check)], '')
+                # replace only the clean_partial value when there are multiple '}'
+                return_value = return_value.replace(return_value[:return_value.find(check) + len(check)], '', 1)
                 match = self.__find_match(return_value[:return_value.find(check) + len(check)])
```
navvrnd: This whole block (if match is None) looks unnecessary code. Can you please check why this is necessary? If not necessary remove and that would solve the original issue as well.

sugan19: It is necessary in cases like, if we have "ab {abc} cd ${list.value}". So we cant remove those lines.

kaheichan:
Ran the test and verified it capture the response correctly

bjrr:
@sugan19  approved by Ka, Venkat, Hari & Arvind. Need merge conflict resolved in order to merge

bjrr:
@sugan19 Why dismiss the reviews from Ka, Arvind, Hari and venkat when they all approved?

bjrr:
This fix is not needed. User can provide his own closing and opening patterns and not use the default to avoid this issue. Closing this issue

PR #223, opened 2017-09-29
WAR-1168: Option to use data repository values in Input Data and Var config files

venkat1077:
Support using data_repo values in IDF & VC files

Logs and test instructions in the Jira ticket.


kaheichan:
Ran the testcase, confirmed it pass on local
also verify it covers the use case
Good job @venkat1077

PR #173, opened 2017-09-13
WAR-1402 Response reference is not saved properly while executing on a different system which is given in the td file.

sugan19:
Fetching the system name and session name from details dict if available, else takes from test case. 
If session name given along with system name then it is split and saved respectively.
Fetches the title and row value, checks if already available in td_resp_dict and updates the response ref key and its value to td_resp_dict else updates the title and row value and then updates the response ref key and its value to td_resp_dict.

While working in WAR-1338, I found this issue. The sample file for this is attached in jira.

PR #229, opened 2017-10-04
WAR-1402 Response reference is not saved properly while executing on a different system which is given in the td file.

sugan19:
Fetching the system name and session name from details dict if available, else takes from test case.
If session name given along with system name then it is split and saved respectively.
Fetches the title and row value, checks if already available in td_resp_dict and updates the response ref key and its value to td_resp_dict else updates the title and row value and then updates the response ref key and its value to td_resp_dict.

While working in WAR-1338, I found this issue. The sample file for this is attached in jira.


warrior/Actions/CliActions/cli_actions.py
```diff
@@ -816,25 +817,21 @@ def send_testdata_command_kw(self, system_name, session_name=None, wdesc='', var
...
-        testdata, varconfigfile = Utils.data_Utils.get_td_vc(self.datafile, system_name, td_tag, vc_tag)
+        testdata, varconfigfile = Utils.data_Utils.get_td_vc(self.datafile, system_name,
+                                                             td_tag, vc_tag)
         #abspaths = Utils.data_Utils.get_filepath_from_system(self.datafile, system_name,
         #                                                     'testdata', 'variable_config')
         #testdata = abspaths[0]
         #varconfigfile = abspaths[1]
         # td_resp_dict = {}
-        status, resp_dict = cli_Utils.send_commands_from_testdata(testdata, session_object,
+        status, resp_dict, td_resp_dict = cli_Utils.send_commands_from_testdata(testdata, session_object,
```
navvrnd: If you need only td_resp_dict, why getting resp_dict as well?? Why the need for extra return parameter ??

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
```diff
@@ -247,7 +248,59 @@ def send_commands_from_testdata(self, testdatafile, **args):
...
+                        if not WarriorCliClass.cmdprint:
+                            td_resp_dict[title_row].update(resp_key_value_dict)
+                    # if title_row value available in td_resp_dict,
+                    # then it updates the resp_ref key and value to td_resp_dict
+                    else:
+                        resp_key_value_dict = {value: temp_resp_dict[value]}
+                        if not WarriorCliClass.cmdprint:
+                            td_resp_dict[title_row].update(resp_key_value_dict)
+
+        return finalresult, responses_dict, td_resp_dict
```
navvrnd: This function has become quite long (167-303 = 137 lines). Can this be refactored to multiple functions? Also use list comprehensions whenever possible to give a readable code in lesser number of lines.

warrior/Actions/CliActions/cli_actions.py
```diff
@@ -860,8 +857,9 @@ def set_session_timeout(self, system_name, session_name=None, int_timeout=30):
 
         if session_object.isalive():
             session_object.timeout = int_timeout
-            Utils.testcase_Utils.pNote("Timeout value is set to {0}mins for the session with system name : {1},"\
-                                       " session name : {2}".format(int_timeout, system_name, session_name))
+            Utils.testcase_Utils.pNote("Timeout value is set to {0}mins for the session with" \
+                                       " system name : {1}, session name : {2}" \
+                                       .format(int_timeout, system_name, session_name))
```
navvrnd: EOL slashes are unnecessary within the parenthesis, please remove them.

sugan19:
#173 
@venkat1077 Addressed your comments
@brettleehari Addressed your comments
@navvrnd line 836 to 842 is together combined as a single if... elif... block.

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
```diff
@@ -247,7 +249,84 @@ def send_commands_from_testdata(self, testdatafile, **args):
...
+                    if len(temp_sys) > 1:
+                        temp_sess_name = temp_sys[1]
+                    else:
+                        temp_sess_name = td_session_list[count]
+                    # session id is formed from the system name and session name.
+                    td_session_id = Utils.data_Utils.get_session_id(temp_sys[0], temp_sess_name)
+                    td_resp_dict = get_object_from_datarepository(str(td_session_id)+"_td_response")
+
+                    # checks if title_row value is not in td_resp_dict
+                    if title_row not in td_resp_dict.keys():
```
navvrnd: .keys() is unnecessary, please remove it.

bjrr:
@navvrnd Please verify that the testcase created to test this functionality covers the feature well and is added to our regression. Please let me know.

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
```diff
@@ -247,7 +249,84 @@ def send_commands_from_testdata(self, testdatafile, **args):
...
+                        # after updating title_row value, it updates the resp_ref key and value
+                        td_resp_dict = append_resp_ref_key_value(value, temp_resp_dict,
+                                                                 td_resp_dict, title_row)
+
+                    # if title_row value available in td_resp_dict,
+                    # then it updates the resp_ref key and value to td_resp_dict
+                    else:
+                        td_resp_dict = append_resp_ref_key_value(value, temp_resp_dict,
+                                                                 td_resp_dict, title_row)
+        return td_resp_dict
```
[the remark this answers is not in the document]
sugan19: @navvrnd Addressed all your comments.

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
```diff
@@ -248,7 +252,85 @@ def send_commands_from_testdata(self, testdatafile, **args):
...
+
+        if finalresult:
+            if len(resp_key_list) != len(td_sys_list):
+                for j in resp_key_list:
+                    td_sys_list, td_session_list = \
+                     get_sys_list_session_list(details_dict, td_sys_list, td_session_list)
+            else:
+                td_sys_list, td_session_list = \
+                  get_sys_list_session_list(details_dict, td_sys_list, td_session_list)
+            for title_row, temp_resp_dict in responses_dict.iteritems():
```
navvrnd: multiple response references in a single cmd should map to the same system, hence this logic should be revisited.

sugan19:
@navvrnd Addressed your review comments

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
```diff
@@ -227,10 +229,30 @@ def send_commands_from_testdata(self, testdatafile, **args):
...
-                    result = (result and rspRes) if "ERROR" not in (
-                                result, rspRes) else "ERROR"
+                    rspRes, resp_key_list = new_obj_session._get_response_dict(
+                        details_dict, i, response, resp_key_list)
+
+                    if resp_key_list:
+                        for resp in resp_key_list[i].keys():
+                            # session id is formed from the system name and session name.
+                            session_id = \
+                             self.get_session_id_for_resp_ref(details_dict, resp_key_list[i],
```
satyakrish: @sugan19 
instead of writing a new method to get the session id that was used to send the command please update the existing method _get_obj_session to get the session_id in addition to the session_object.
Because the _get_obj_session is used to form the session_id to which we send the command but today it return the session_object and not the session_id.. but it does have the session_id information.
So please update _get_obj_session and wile updating do not modify the existing return statement but add an additional flag to request sesssion_id to be returned and if the flag is set add a new return statement to return session_id as well.

We should not be forming session_id's in two places , it will be difficult to track changes.. so it is best to update the method that forms the session_id to which the command is sent to...

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
```diff
@@ -227,10 +229,30 @@ def send_commands_from_testdata(self, testdatafile, **args):
...
+                                                              resp, i, system_name,
+                                                              session_name, key)
+                            td_resp_dict = get_object_from_datarepository(str(session_id))
+                            # checks if title_row value is not in td_resp_dict
```
satyakrish: I'm finding it difficult to understand the flow here, can you please explain the flow of what is being done after getting the session-id

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
sugan19: @satyakrish Addressed you review comments

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
sugan19: In line 271 it checks for title_row value not in td_resp_dict, 
Consider a td block which has 3 commands: 
td1 --> c1, c2, c3 and its respective sys tag s1, s2, s1

In the above scenario, if the response reference is stored for td1_c1 then it need to check the particular session exists. After that it needs to check if the title/row is present, if not it needs to be added to the td_resp_dict. 
The same is done for td1_c2 since it has a different sys 's2'.
In td1_c3 it will check for the title/row value which would be available since the title/row was updated to the td_resp_dict of sys 's1' for 'c1' command

satyakrish: Thank you for the explanation .. good one.

I have some questions the way td_resp dict is handled, please go through the scenario below and can you add a test to make sure it works fine.

td1  : 
cmd1   resp_req = y (default)
cmd2   resp_req = n
cmd3   resp_req = y (default)
cmd4   resp_req = n 

First command:
- on executing the first command, a empty td_resp_dict is sent to 'update_resp_ref_to_repo'
- since resp_req = y, you will grab the td_resp_dict from data repo, update it and return it back. 
I assume the dictionary is passed by reference it should update both data repo and funciton local copy as well But we need to have a test case to test this.

continue.. 
second command:
- for the second command you will send the updated td_resp_dict to 'update_resp_ref_to_repo'

- Since resp_ref=n you will return whatever was received.

third command:
- send updated td_resp_dict to 'update_resp_ref_to_repo' 
- since resp_ref = y you will grab the td_resp_dict from data_repo .. we need to make sure that the update we made for command one in line-277 actuall updated the data repo as well.
Because if it did not then in tis case we will lose command-1 response..

Can you please add a test with this scenario and make sure all command responses are updated in data_repo and in function local copy as well

sugan19:
Hi @satyakrish,

Point#4 - Added the new scenario to warrior_fnc_tests/warrior_tests/testcases/cli_tests/other_attr/tc_different_sys_single_resp.xml and the command gets updated to both data repo and funciton local copy

Addressed all your comments

satyakrish:
approving code changes as the regression test cahnges will be tracked in the respective repo

PR #230, opened 2017-10-04
WAR-1445: Iteration by test block not working for list

sugan19:
If list is given in per_td_block, then it should be iterated.
Currently all values in list will be printed next to next.

Sample file is attached in the ticket.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,73 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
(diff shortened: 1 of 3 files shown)


warrior/Framework/ClassUtils/testdata_class.py
```diff
@@ -555,7 +566,19 @@ def resolve_iteration_patterns(self, details_dict):
...
+                if value not in cmd and '${' in cmd:
+                    t_obj = TestData()
+                    t_obj.list_substitution_precheck(vc_file_list[i], details_dict, '${', '}')
+                    cmd_value = t_obj.cmd_sub(details_dict, i, vc_file_list[i], '${', '}')
+                    if cmd_value[0] is not False:
+                        t_obj.align_cmd(details_dict, i, cmd_value)
+                        for param in CMD_PARAMS.keys():
+                            del details_dict[param][i]
+            else:
+                value = cmd.split('+')[0]
```
venkat1077: Can you check if this logic can be done along with '+' iteration/td_block logic??

navvrnd:
From Offshore, Venkat is reviewing.

PR #232, opened 2017-10-04
WAR-1447 : Adding option in set env kw to choose from terminal too

mohfuj:
As a user I want options in set env KW. 

Option is to chose from terminal if it is available.

### DIFF
--- a/warrior/Actions/CommonActions/common_actions.py
+++ b/warrior/Actions/CommonActions/common_actions.py
@@ -172,29 +172,34 @@ class CommonActions(object):
             return False
 
     def set_env_var(self, var_key=None, var_value=None, filepath=None,
-                    jsonkey="environmental_variables"):
-        """create a temp environment variable
-        the value will only stay for this run
+                    jsonkey="environmental_variables", overwrite = "yes"):
+        """Create a temp environment variable, the value will only stay for the current Execution
         :Argument:
             var_key = key of the environment variable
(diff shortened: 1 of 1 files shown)


warrior/Actions/CommonActions/common_actions.py
```diff
@@ -172,29 +172,34 @@ def verify_data(self, expected, object_key, type='str', comparison='eq'):
...
         if not any([var_key, var_value, filepath]):
-            print_error('Either Provide values to arguments \"var_key\" & \"var_value\" or to argument \"filepath\"')
-
-        if var_key is not None and var_value is not None:
+            print_error('Either Provide values to arguments \"var_key\" & \"var_value\" or to '
+                        'argument \"filepath\"')
+        if overwrite == "NO" and os.getenv(var_key):
+            print_info("Using ENV variable {0} set earlier with value '{1}'".format(var_key,
+                                                                           os.getenv(var_key)))
+        if var_key is not None and var_value is not None and overwrite == "YES":
```
venkat1077: this condition should work for overwrite == "NO" condition as well.
Use-case - when overwrite is "NO" and the value is not set earlier.

bjrr:
@venkat1077 @brettleehari The ticket is 1447 not 1147. Please always verify the ticket number before approval. I made the change in order to merge based on your approvals.

PR #236, opened 2017-10-05
WAR-1442: Print statement needs to be improved for verify_data and store_in_repo keywords

sugan19:
currently store_in_repo and verify_data does not give any info about what are the keys and its value which are executed, so printed the details.

In verify_data keyword, if title is available and key does not exist then it prints only the exception which the user may not understand, so printed a warning message.

Sample testcase attached in the ticket.


warrior/Framework/Utils/data_Utils.py
```diff
@@ -901,10 +899,15 @@ def validate():
     keys = key.split('.')
     value = get_object_from_datarepository(keys[0])
     for k in keys[1:]:
+        if k not in value:
+            print_warning("key {} not present in data repository\n".format(key))
         value = value[k]
     if not value:
         err_msg += "key {} not present in data repository\n".format(key)
         result = "ERROR"
+    if value:
```
sanikakulkarni: Hi @sugan19,

Since we have the ```if not value``` block on line 905, can we just have an ```else``` instead of ```if value``` on line 908?

Thanks.

warrior/Framework/Utils/data_Utils.py
```diff
@@ -901,10 +899,15 @@ def validate():
     keys = key.split('.')
     value = get_object_from_datarepository(keys[0])
     for k in keys[1:]:
+        if k not in value:
+            print_warning("key {} not present in data repository\n".format(key))
         value = value[k]
```
sanikakulkarni: Hi @sugan19,

In line 902, you have added an ```if``` statement to check if k does not exist in value. If it does not, then you are printing a warning - which works.

Now, once we exit that ```if``` block, we reach line 904, where we are assigning ```value``` the value of ```value[k]```; but if ```k``` does not exist in ```value```, line 904 will throw an error. 

I would suggest that the statement ```value = value[k]``` be put inside an ```else``` block so that this situation is avoided.

Thanks.

warrior/Framework/Utils/data_Utils.py
sugan19: @sanikakulkarni  Addressed you comments.

warrior/Framework/Utils/data_Utils.py
sugan19: @sanikakulkarni  Addressed you comments.

sanikakulkarni:
Hi @sugan19,

Please attach the CLI regression logs to the jira ticket. If you are unsure as to where to find the CLI project, please get in touch with @venkat1077.

warrior/Framework/Utils/data_Utils.py
```diff
@@ -900,15 +897,29 @@ def validate():
...
+                result = "FALSE"
+                print_warning("The key, value pair '{0}:{1}' present in the  "
+                              "data_repository doesn't satisfy the expected value & "
+                              "condition '{2}:{3}'".format(key, value, comparison, expected))
+            else:
+                print_info("The key, value pair '{0}:{1}' present in the  "
+                           "data_repository satisfies the expected value & condition "
+                           "'{2}:{3}'".format(key, value, comparison, expected))
+        except KeyError:
+            print_error(key_err_msg.format(key))
```
sanikakulkarni: Hi @sugan19,

Please catch a general exception and print it out here. You can refer to line 879 of this file.

warrior/Actions/CommonActions/common_actions.py
```diff
@@ -165,11 +173,13 @@ def verify_data(self, expected, object_key, type='str', comparison='eq'):
                                                      type, comparison)
         if result == "FALSE":
             print_error("Expected: {0} {1} {2} but found {0}={3}".format(
-                object_key, comparison, expected, value))
+                        object_key, comparison, expected, value))
         if result == "TRUE":
             return True
-        else:
+        elif result == "FALSE":
```
sanikakulkarni: Hi @sugan19 ,

the ```result == 'FALSE'``` block is repeated on line 174 and 179. Please merge those blocks.

warrior/Framework/Utils/data_Utils.py
```diff
@@ -900,15 +897,29 @@ def validate():
...
+        try:
+            for k in keys[1:]:
+                value = value[k]
+            if result == "ERROR" or result == "EXCEPTION":
+                print_error(err_msg)
+            elif not comp_funcs[comparison](value, exp):
+                result = "FALSE"
+                print_warning("The key, value pair '{0}:{1}' present in the  "
+                              "data_repository doesn't satisfy the expected value & "
+                              "condition '{2}:{3}'".format(key, value, comparison, expected))
```
sanikakulkarni: Hi @sugan19,

In line 911, ```format(key, value, comparison, expected)``` would print out the ```key``` argument rather than ```k``` from the for loop in line 903 - was this intentional?

Thanks.

warrior/Framework/Utils/data_Utils.py
```diff
@@ -900,15 +897,29 @@ def validate():
...
+            else:
+                print_info("The key, value pair '{0}:{1}' present in the  "
+                           "data_repository satisfies the expected value & condition "
+                           "'{2}:{3}'".format(key, value, comparison, expected))
```
sanikakulkarni: Hi @sugan19,

In line 915, ```format(key, value, comparison, expected)``` would print out the ```key``` argument rather than ```k``` from the for loop in line 903 - was this intentional?

Thanks.

warrior/Framework/Utils/data_Utils.py
sugan19: @sanikakulkarni Addressed your comments

warrior/Actions/CommonActions/common_actions.py
sugan19: @sanikakulkarni Addressed your comments

warrior/Framework/Utils/data_Utils.py
sugan19: @sanikakulkarni Yes, this was intentional. The 'key' value has the entire key (eg: if user gives k1.k2.k3) but k has only k1/k2/k3.

warrior/Framework/Utils/data_Utils.py
sugan19: @sanikakulkarni Yes, this was intentional. The 'key' value has the entire key (eg: if user gives k1.k2.k3) but k has only k1/k2/k3.

sanikakulkarni: This print statement is good.

warrior/Framework/Utils/data_Utils.py
sanikakulkarni: Hi @sugan19,

Since this print statement prints out a negative scenario, it would be best if the entire key ( key ) is printed out as well as specifically ( k ) is also printed out.

Consider the scenarios:

1. In the current print statement, the user has no way to know was missing (since k is not being printed separately)
2. If you just print out "k", user may not be able to figure out which "k" is missing as "key" may have duplicate "k"s

Hence, printing "key" and "k" would be the most beneficial

Thanks.

sanikakulkarni:
Hi @sugan19,

I ran a custom testcase and got this error:

```-E- Expected: demovar.demochild.demograndchild eq 1 but found demovar.demochild.demograndchild=1```

Now, this error happened because in the datarepository, the ```1``` that was stored was an integer while verifying the this data the ```1``` that was supplied was a string. Hence, the mismatch which caused the error. Please add a print staement that also states the data type of each of the value.

Thanks.

warrior/Framework/Utils/data_Utils.py
sanikakulkarni: Hi @sugan19,

I am not sure if I understand what you mean by ```The print_warning in line: 909 is executed only if it has the entire key and k has no scope in elif block.```. Can you please elaborate?

Thanks.

sanikakulkarni:
Hi @sugan19,

I tested the following scenarios:

Positive:
1. Step 1: ```store_in_repo``` keyword with arguments ```datavar = demovar``` and ```datavalue = demovalue```. Step 2: ```veify_data``` keyword with arguments ```object_key = demovar```, ```expected = demovalue ```, and ```comparison = eq ``` - **PASS**
2. Step 1: ```store_in_repo``` keyword with arguments ```datavar = demovar.demochild``` and ```datavalue = demovalue```. Step 2: ```veify_data``` keyword with arguments ```object_key = demovar.demochild```, ```expected = demovalue ```, and ```comparison = eq ``` - **PASS**
3. Step 1: ```store_in_repo``` keyword with arguments ```datavar = demovar.demochild```, ```datavalue = 1```, and ```type = int ```. Step 2: ```veify_data``` keyword with arguments ```object_key = demovar.demochild```, ```expected = demovalue ```, ```type = int ```, and ```comparison = eq ``` - **PASS**

sugan19:
Hi @sanikakulkarni,

I have addressed all your comments and added a test to our Travis CI regression with all scenarios. Kindly review it and let me if there are further comments. 

I have initiated regression for this PR will attach in ticket once it is done.

Test: wftests/warrior_tests/testcases/common_actions_tests/verify_data_test.xml

sugan19:
Hi @sanikakulkarni 

I have attached all regression logs to the ticket. Kindly verify it

PR #237, opened 2017-10-05
bugfix/war-1444: code to checkout warriorspace branches in warhorn

navvrnd:
label tag for warriorspace does not work in warhorn which needs to be fixed. 

Testing: create default_config file with branch provided as label and check that the warriorspace is cloned and copied with the right branch checked out.

### DIFF
--- a/warhorn/warhorn.py
+++ b/warhorn/warhorn.py
@@ -231,11 +231,11 @@ def print_out_checkout_status(label, check, current_tag, repo_name, **kwargs):
     """Git checkout status reporting"""
     logfile = kwargs.get("logfile")
     print_log_name = kwargs.get("print_log_name")
-    if check and current_tag == "":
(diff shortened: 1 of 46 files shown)


warhorn/warhorn.py
```diff
@@ -780,14 +780,14 @@ def clone_warriorspace(base_path, current_dir, **kwargs):
     logfile = kwargs.get("logfile")
     config_file_name = kwargs.get("config_file_name")
     print_log_name = kwargs.get("print_log_name")
-    internal_copy = kwargs.get("dest")
     node = get_node(config_file_name, "warriorspace")
     if node is False:
-        print_error("Drivers tag not found. No drivers cloned.",
+        print_error("warriorspace tag not found. No warriorspace cloned.",
```
brettleehari: Good find!

PR #238, opened 2017-10-05
WAR-1450: Print the 'expected result' before the test case starts

venkat1077:
Code change(s): Print the value provided in the 'Expected Results' field of a testcase during execution.

How to test: Run any test to see if the value provided in the 'Expected Results' field is being printed in the console along with other testcase details.
Logs: Regression logs are attached in the Jira ticket


bjrr:
@venkat1077  The pull request implements the expected results. However, the Jira ticket states Test Case description, steps and expected result. Did we decide that we no longer need the Test Case description, and steps? If so can you please update the Jira ticket with what is agreed on.

PR #245, opened 2017-10-11
WAR-1479 Support multiple rules/exectypes for hybrid mode

kaheichan:
Add exec_type_driver to hybrid driver
so it can process rules in execute tag first before step execution

Test log are attached in Jira

### DIFF
--- a/warrior/WarriorCore/Classes/hybrid_driver_class.py
+++ b/warrior/WarriorCore/Classes/hybrid_driver_class.py
@@ -68,6 +68,7 @@ from Framework.Utils.testcase_Utils import pNote
 from Framework.Utils.print_Utils import print_debug, print_info
 from WarriorCore import step_driver, common_execution_utils
 import WarriorCore.onerror_driver as onerror_driver
+import WarriorCore.exec_type_driver as exec_type_driver
 
 
 class HybridDriver(object):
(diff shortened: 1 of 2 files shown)


venkat1077:
@kaheichan Code looks good. Just a single query, I have added it in Jira ticket.

warrior/WarriorCore/Classes/hybrid_driver_class.py
```diff
@@ -213,6 +214,36 @@ def _execute_step_list(self):
...
                 break
 
+            # Decide whether or not to execute keyword
+            # First decide if this step should be executed in this iteration
+            if not goto_stepnum or goto_stepnum == str(step_num):
+                # get Exectype information
+                run_current_step, trigger_action = exec_type_driver.main(step)
+                if not run_current_step:
+
+                    if trigger_action.upper() in ['ABORT', 'ABORT_AS_ERROR']:
```
satyakrish: if the action is abort, stop_after_current_execution and stop_after_current_step should be set only if the iteration type is not standard.. this is because for standard iterative behavior when the action is abort we only abort the current iteration and move on to the next iteration.

satyakrish: if the action is abort, stop_after_current_execution and stop_after_current_step should be set only if the iteration type is not standard.. this is because for standard iterative behavior when the action is abort we only abort the current iteration and move on to the next iteration.

kaheichan: Addressed this comment in the 8b18946d82f5bf30e28e989dcac1c5efb160fe89

PR #251, opened 2017-10-13
Make warrior a package. Run setup.py to install

kamranHusain:
Do not use WAR-210;  Use this instead.

### DIFF
--- a/warrior/__init__.py
+++ b/warrior/__init__.py
@@ -1,0 +1,24 @@
+'''
+Copyright 2017, Fujitsu Network Communications, Inc.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


kamranHusain:
Run setup.py to install

kamranHusain:
Here is a sample program to use. 

 
import os, sys
import warrior
import warrior.Framework as Framework
import warrior.Tools as Tools
import warrior.Framework.ClassUtils as ClassUtils
from Framework.Utils import file_Utils 

print "Testing ...", 
filename = '/home/khusain/Projects/forDemo/warriorframework/AUTHORS'

xstr = "ack"
print "Searching ", filename , " for " , xstr 
print file_Utils.searchFile(filename,xstr) 
print file_Utils.getFileName(filename) 

Here is the output from it. 

Testing ... Searching  /home/khusain/Projects/forDemo/warriorframework/AUTHORS  for  ack
True
AUTHORS

PR #260, opened 2017-10-17
War-1097: Support for * in path to loop through all project and suite files under directory

kamranHusain:
I only modified two files: project_driver.py and testsuite_driver.py in Warrior Core.


bjrr:
@kamranHusain @kaheichan @sanikakulkarni @satyakrish This pull request has 601 files changed. This should only be a few files.

kamranHusain:
Created WAR-1097x instead of WAR-1097 branch because the original branch was from katana-develop

PR #261, opened 2017-10-17
WAR-1396 Add logging support for core apps as an example

kamranHusain:
Only three apps are shown as an example. Developers must add their own apps to the settings.py file.


sanikakulkarni:
Closing as this would be reworked

PR #268, opened 2017-10-24
WAR-1097 Support for the asterisk is suites and projects

kamranHusain:


satyakrish:
@kamranHusain 
cc: @bjrr 
This pull request is not valid the tetsuite driver and other files are outdated, you need to get the latest develop, and submit your changes using that latest develop.

PR #270, opened 2017-10-26
WAR-912 : Multiple conditional statements for Execute Type

priyaarun23:
Testcase  --> Step --> Execute Type : 
     >For the Execute Type 'If' & 'If Not', Multiple conditional statements are supported.
     >Operator has been added for each rule which is an optional field.
     >Top level Else & Else Value for the Execute Type in case the expression fails.
     >User provided Expression field for the logical operations among the rules.

### DIFF
--- a/katana/static/app/controllers/testcaseCapCtrl.js
+++ b/katana/static/app/controllers/testcaseCapCtrl.js
@@ -30,7 +30,6 @@ app.controller('TestcaseCapCtrl', ['$scope','$routeParams','$http', '$location',
         $scope.step_onerror = "next";
         $scope.step_onerror_value = "";
         $scope.arg_list = [{"_name": "", "_value": ""}];
-        //$scope.rule_list = [{"_Condition": "", "_Operator": "eq", "_Condvalue": "", "_Else": "next" , "_Elsevalue": ""}];
         $scope.showStepEdit = false;
         $scope.insertStep = false;
         $scope.alldirinfo = "";
@@ -43,10 +42,8 @@ app.controller('TestcaseCapCtrl', ['$scope','$routeParams','$http', '$location',
         $scope.showRulesBelow = false;
         $scope.hideExp = true;
         $scope.hideElse = true;
-       // $scope.hideElseVal = true;
         $scope.ExecTypeVal = 0;
-        // $scope.ruleTable = false;
-       
+        $scope.editstepcheck = 0;
 
         function readConfig(){
             getConfigFactory.readconfig()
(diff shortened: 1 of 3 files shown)


kaheichan:
Hi @priyaarun23 
is this ready to be reviewed again?

priyaarun23:
@kaheichan 
First issue that you mentioned in JIRA ticket is the existing issue in Latest Github. Created new ticket WAR-1550 for the same.
Second issue in editing a step of existing TC file (multi_rules_with_expressions.xml) has been addressed.

Auto population of default value for Runmode has been handled as 'Standard' since the xml file has <rmt/> tag for it.

Know issue : As we don't have values for '_Else_' and '_operator_' in the XML file(multi_rules_with_expressions.xml), onload of a page show empty value. When you edit and save, the default values have been updated. 
Tags similar to <rmt/> could be added in XML. If not, need to push the default values to all the empty tags from katana side.


Thanks,
Priya (Nov 01)

priyaarun23:
Hi @bjrr @kaheichan ,

Addressed the review comments as below :
     1-Issue described by @kaheichan in JIRA(WAR-912) has been addressed. All steps could be rendered/edited/updated with relevant data.
     2-Deleted the logic of Else & Else Value in Rules. Now we have only one Else in Expression(global) level.
     3.Changed the Expression label on Katana (xml tag remains the same) as Multi Rule Expression.
     4.Haven't received mail for tooltip so it is untouched.


Thanks,
Priya (Nov 02)

kaheichan:
Hi @priyaarun23 
Small comments, 
1. The text "Multi Rule Expression" isn't showing in full in katana. Please make the box larger to show the full text
2. When opening file with rules that doesn't have an operator, the first 2 rules in the step won't show the default operator value eq. This only happens when user edit the step. Please fix it to correctly show the default value

priyaarun23:
Hi @kaheichan ,

     1.The Label width that you fixed for 'Multi Rule Expression' breaks its 'Else Value' alignment. Hence I have changed the layout just like other fields of TC->Step(Ex : On Error -- Goto). Attached the screenshots for the same
multiexpression_break
multiexpression_new layout

.
     2 & 3 Handled.

priyaarun23:
multiexpression_break
multiexpression_new layout

priyaarun23:
@kaheichan ,
Request you to review the changes.

sanikakulkarni:
Pull Request has been approved by @brettleehari which was later dismissed on account of new commits.

kaheichan:
Verified UI component are aligned and functioning
Created a new testcase file with Multiple rules, expressions and goto else value
Check in the xml that value are created
Run the created testcase without problem

PR #272, opened 2017-10-26
War-1518: RMT was not iterating on suite files

brettleehari:
RMT is not working on Suite files. Identified the bug and resolved the same. 
For RMT the attribute is Number_Attempts and For RUP/RUF the attribute is MAX_Attempts. 
But when user does not give MAX_Attempts for RMT(which is not required for RMT) Test fails.
This fix address it. 

A Test with appropriate value is added in the WFTEST and logs are available on the ticket.


kaheichan:
Hi @brettleehari 
Looks like the test for this bugfix isn't an automated test
Please create a project that properly test how many times each RMT/RUP/RUF suite has run
A sample implementation would be
Project
-Suite A RMT 3
--Testcase A
---step increase_value key_name
-Suite B RUP 5
--Testcase B
---step increase_value key_name_b status FAIL max_value 3 max_status PASS
-Suite C RUF 5
--Testcase C
---step increase_value key_name_c status PASS max_value 3 max_status FAIL
set Suite C context as negative as its final status will be FAIL and will be flipped by project_driver
-Suite D
--Testcase D
---step verify_data key_name 3
---step verify_data key_name_b 3
---step verify_data key_name_c 3

brettleehari:
@kaheichan I beg to differ to the comments above. 
The testcase I added is to verify the bug when RMT is used and MAX attempt is not provided. 
The testcase added fails in the develop branch, which means the new test added captures the issue reported. 
Moreover the testcase added passes in the WAR-1518 branch, which means the testcase fix is available in the branch. 
My understanding is that for this bug fix the added test suffice as the ticket tries to fix a logic of MAX attempts in RMT and not on the number of times RMT/RUF/RUP ran. 

Your suggestion is perfect for WAR-1348. I would take up your suggestion for WAR-1348. What do you think. Since the issue reported and fix submitted is not about how many times RMT ran but about RMT requiring an attribute MAXATTEMPTS which has no logic in runmode RMT. 
Request you to reconsider your change request.

kaheichan:
Hi @brettleehari 
Agree that RUP/RUF is out of scope of this PR
but for RMT, a simple passing test isn't sufficient in this case
If RMT doesn't work, the said test will still pass
In general, in order to test runmode functionality, having a verify step is a good way to go
It will definitely catch what the current test can't catch

The idea of testing isn't just to test the fix, we should strive for the best when we see there's a chance of improvement, and in this case it is a perfect chance to enhance our tests

brettleehari:
Dear @bjrr,
                        Can we have a call to dicsuss this item. I firmly believe the test added is a fit case for the bug. I humbly disagree with @kaheichan 's argument. 
To give you a brief, the bug and fix was not on number of times each step is iterated/run. 

The bug is simple, RMT requires number_attempts argument but current logic for RMT is expecting the argument max_attempts and only in its absence number_attempts is consider. 
Test added covers the above logic. Please tell me if the test does not cover the fixed logic, which is not RMT core logic but parses the correct argument from the user.

In Warrior we should encourage unit test rather than functionality test. With functionality test the scope to test is limited and coverage is minimum. Moreover Tests & coverage are a subjective item can we take a firm stand and reject code fix because we have not strived for the best when we see there's a chance of improvement? Isn't this subjective ? The feature (RMT) exists for so long, a lot of improvements commits pullrquest have gone into the feature but we do not have a single test to test RMT's core functionality ?
Given that this is the state why should we reject this particular case (which is actually not fixing any logic in RMT core functionality) this bug just parses the correct user information.  

Kindly request you to point me to a test which will fail if RMT does not work? 
Maybe i could leverage an existing test. The test suggested is an overkill for this simple bug. This code fix is not introducing any logical to RMT core functionality. 
 
With functionality test each reviewer will have his/her own view and is purely subjective some may be very aspirational and some may be conservative. So, I propose a solution and not only bring issue to table. 
Tests Proposal for Bugs: 
If the new test catches the existing bug in the current code then it is adequate any test above the adequate is a bonus and not a requirement. Fix should not be rejected based on bonus but on adequacy.

kaheichan:
Approving after tests are added

PR #284, opened 2017-10-27
war-1507: result from _get_response_dict is not combined properly with the result from _send_command_retrials

navvrnd:
failure in variable substitution in testdatafile does not result in ERROR status for the step.

### DIFF
--- a/warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
+++ b/warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
@@ -235,7 +235,8 @@ class WarriorCli(object):
                         response=response, system_name=td_sys)
                     rspRes, response_dict = new_obj_session._get_response_dict(
                         details_dict, i, response, response_dict)
-                    result = result and rspRes
+                    result = result and rspRes if "ERROR" not in (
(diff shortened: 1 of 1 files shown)


kaheichan:
Hi @navvrnd 
Please update the ticket title to truly reflect what this bugfix is about
The bug is in send_commands_from_testdata 
when the result of _send_cmd_get_status, _send_command_retrials and _get_response_dict result are combined incorrectly, ignoring any "ERROR" status coming from any of these functions

variable substitution error is one of the scenario that any of these functions can fail, any error from these functions will encounter this bug

navvrnd:
@kaheichan updated the ticket title. Please let me know, if anything else is pending from my side.

navvrnd:
Hi @kaheichan Why can't this be pushed for this release? Please let me know, if there is any thing pending from my side.

kaheichan:
Hi @navvrnd 
just put this in the release as this is created after the sprint planning
made a comment in internal PR, once that is addressed this can be merged

navvrnd:
Thanks @bjrr . @kaheichan, Brian has put this in release. Please take this for the release.

kaheichan:
Code looks good, attached the latest regression

PR #291, opened 2017-10-31
WAR-1046 : Introducing interval for RUNMODE at Testcase level

brettleehari:
Changes:
Updated the common utils function to parse interval from runmode. Returns three values instead of two.
Wherever we called the function, added recipient variables to match the return values.3 instead of 2.
Added logic to wait for a particular amount of time for Runmode in cases.
Moved the wait logic from Common Actions to date time utils, such that it could reused for all wait/sleep/timeout

Attached regression logs to WAR ticket. 
Brief : 
Like WAR-1348, this feature can be implemented across cases, Suites & Project.
The current implementation only for Cases, because WAR-1348 attempts to modularise most of the run mode code. When implemented on top on WAR-1348 we will have an interval for RUN MODE in a much-refined manner. 

This logic was previously on top of WAR-1348, but since WAR-1348 is not going in adding a simpler code change to unblock user community.


brettleehari:
Hi @kaheichan,
                          Updated tests as suggested! 
Regarding the design, the design is dependent WAR-1348 which is placed on hold at the moment. Have created a WAR ticket to introduce interval at all levels of Warrior such as cases/suites/Project. So the design document would be updated in WAR-1544 based on WAR-1348. 
Current design is simple, if interval optional field is available will wait for seconds present in the runmode before executing the next step.

kaheichan:
Improve test to cover all intervals between runmode execution
explain how test functioned in keyword description
change interval (wait_time/timer) to runmode_timer as this matches with cli retry timer
update testcase file to use updated keyword

Run regression and new test, status is pass
Not able to review design as design doc is not updated and not all code change documented in the current doc, only verify functionality

PR #292, opened 2017-10-31
WAR-1545 test for RMT on suite detail level

kaheichan:
Include changes in WAR-1518 for testing


kaheichan:
Test files added to WAR-1518

PR #298, opened 2017-11-03
Release warrior 3.5.0

kaheichan:


kaheichan:
Failure is caused by pylint not ran when snmp related PR is merged from forked repo into develop
so it didn't catch the degrade until merging develop into master

PR #303, opened 2017-11-06
Release warrior 3.5.0

kaheichan:


warrior/Warrior
```diff
@@ -17,15 +17,30 @@ if sys.version_info < (2, 7):
     sys.exit(0)
 try:
     import os
+    print "import os was successful"
     import shutil
+    print "import shutil was successful"
     import Framework.Utils.email_utils as email
+    print "import email was successful"
     import Framework.Utils as Utils
+    print "import Utils was successful"
```
brettleehari: Do we need these prints ?

PR #304, opened 2017-11-07
Added hook for context switch

kamranHusain:


katana/wui/core/static/core/js/main.js
```diff
@@ -132,6 +132,16 @@ var katana = {
 		$elem.siblings().removeClass('active');
 		$elem.addClass('active');
 		katana.$activeTab = katana.$view.find( '#' + uid ).removeClass('hidden');
+		var onSwitch = katana.$activeTab.find('[on-switch]');
```
keenan35i: all of this can be replaced with katana.methodCaller( 'stringFunction', $elem )

kamranHusain: How does a coder know *when* to call this method?

kamranHusain: Here is a use case.  This will save a developer the onus of (re-)setting the global variable on every function. See the how to use below. 
contextswitch1
contextswitch2

The user defined context switch function can hide popups, set variables or even set a timed function. It just has to be quick.

sanikakulkarni:
Closed because decided that this feature is not required.

PR #313, opened 2017-11-24
WAR-1445: Iteration by test block not working for list

sugan19:
If list is given in per_td_block, then it should be iterated.
Currently all values in list will be printed next to next.

Test is added to warrior_fnc_test

### DIFF
--- a/warrior/Framework/ClassUtils/testdata_class.py
+++ b/warrior/Framework/ClassUtils/testdata_class.py
@@ -562,6 +562,7 @@ class TestDataIterations(object):
         cmd_loc_list = [0]
         cmd_size = 1
         cmd_list_subs = []
+        is_iter = False
         for i, cmd in enumerate(cmd_list):
             vc_file = vc_file_list[i]
             iteration_status = self.validate_iteration_patterns\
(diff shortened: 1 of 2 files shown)


warrior/Framework/ClassUtils/testdata_class.py
```diff
@@ -573,23 +587,31 @@ def resolve_iteration_patterns(self, details_dict):
                 # move on to the next command
                 # This also means that the command parameters do not have any
                 # iteration patterns because the validation was already done.
-                if cmd_iter_pattern != "":
-                    # if cmd_iterpattern is not ""
-                    # call the expand cmd_params method to resolve the
-                    # iteration patterns in the command and get a updated
+                if isinstance(cmd, str) and '+' not in cmd:
+                    # check if list exists in cmd
+                    td_obj = TestData()
```
satyakrish: why intialize a new TestData object here, please use self.list_substitution_precheck

sugan19: @satyakrish 
list_substitution_precheck is in TestData Class and I am using it in TestDataIterations Class, So creating a new object for it.

PR #133, opened 2017-08-14
WAR-1304 : Add the selenium Webdriver and browser version details

brettleehari:
To debug selenium issues, we need the underlying Selenium webdriver version and the browser version.
Launching browser is the entry point in our selenium design, so added two print_info added to cater to this requirement. 

The current change would include the below bold lines
-I- Opens browser instances
-I- 
<< Substep >>
-I- Keyword Description: Opens browser instances
-D- Opening browser 'firefox'
-**I- The Selenium Webdriver version is '2.48.0'
-I- The Browser 'firefox' version is '39.0.3'**
-I- Opening url 'https://www.stackoverflow.com'

PR #62, opened 2017-07-04
WAR-724 - To test if all the available selenium functionality in Warrior is compatible with Selenium 3.0

nalagapp:
1. Added the support for Firefox version above 47 using geckodriver and Selenium 3.x
2. Added the support for Chrome browser using chromedriver

Results files are attached in JIRA ticket.

How-to-test:
Steps are added in user doc and attached in JIRA ticket

PR #333, opened 2017-12-19
WAR-1548 Refactor selenium keyword to support both selenium 2.0 and 3.0 with different version of chrome/firefox

kaheichan:
This ticket also includes changes for WAR-1591
and necessary support for WAR-1074 to run correctly

With firefox version 47 and above, a geckodriver is needed in the local sys env path
https://github.com/mozilla/geckodriver/releases
Chrome needs chromedriver
https://sites.google.com/a/chromium.org/chromedriver/downloads

if geckodriver is needed, then selenium version needs to be 3.5 or above
https://github.com/mozilla/geckodriver#selenium
which means for Firefox 47 and above, warrior needs both geckodriver and selenium >= 3.5 to run
Chrome isn't affected

For geckodriver launched firefox
currently found that browser.capabilities dict has "browserVersion" instead of "version"
needs to handle this kind of key changes in warriorframework

Code is not ready to review, hence no reviewers have been added
PR is opened solely for compare purpose

### DIFF
--- a/warrior/Actions/SeleniumActions/browser_actions.py
+++ b/warrior/Actions/SeleniumActions/browser_actions.py
@@ -215,14 +215,21 @@ class browser_actions(object):
                 browser_details = selenium_Utils.\
                     get_browser_details(browser, datafile=self.datafile, **arguments)
             if browser_details is not None:
-                if type == "firefox":
-                    ff_profile = self.browser_object.\
-                        set_firefoxprofile(proxy_ip, proxy_port)
-                if binary != "" and gecko_path != "":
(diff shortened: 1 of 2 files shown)


kaheichan:
Hi @brettleehari 
As said in the description...this PR is for code comparison purpose
please don't join as reviewer as it is not ready to be reviewed

kaheichan:
This PR is built on top of develop which has #62 and #133 changes
and causes instability to the platform as the code logic is flawed
will create new PR that revert those changes first and provide a stable release

PR #335, opened 2017-12-27
WAR:1583 - Incorrect Results printed on Console window for iterative Parallel TC

venkat1077:
Issue - Duplicate entries in the execution summary when executing 'iterative_parallel' suite.
Reason - Same junit object is being added multiple times to the multiprocessing queue during 'iterative_parallel' suite execution.
Fix - junit_object/python_process is same for all the cases executed in the same system for 'iterative_parallel' suite execution so do not append junit object into queue for all the cases(just add once).  Code changes are in 5a83b8c, rest are pylint fixes.

How to test - Run any suite with exectype as 'ITERATIVE_PARALLEL' with more than one system in a data_file(test use-case and regression results are in the jira ticket).

### DIFF
--- a/warrior/WarriorCore/iterative_parallel_testcase_driver.py
+++ b/warrior/WarriorCore/iterative_parallel_testcase_driver.py
@@ -50,7 +50,8 @@ def execute_iterative_parallel_testcases(system_list, testcase_list, suite_repos
                                   ("auto_defects", auto_defects),
                                   ("system", system),
                                   ("tc_parallel", tc_parallel),
-                                  ("output_q", output_q)
+                                  ("output_q", output_q),
+                                  ("ts_iter", True)
                                   ])
 
         process, jobs_list, output_q = create_and_start_process_with_queue(target_module, tc_args_dict, jobs_list, output_q)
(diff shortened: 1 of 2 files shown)


brettleehari:
Only Ka's review suffice.

kaheichan:
Hi @venkat1077 
I am still seeing duplicate TC result from the Travis log for this ticket
https://travis-ci.org/warriorframework/warriorframework/jobs/322152490 (scroll to the end)
is this the expected behavior?

venkat1077:
@kaheichan 
I'm afraid that this issue is bigger than what I actually thought. I fixed the issue of duplicate entries in results summary when executing a single Iterative_Parallel suite - this works, no more duplicate entries found.

After looking at the shared travis logs, I found that this occurs for normal parallel executions as well.

Use-case: Project(sequential suites) with two suites and the exec_type as "parallel_testcases" for both suites.
Potential reason for the Issue - Same junit object is being used to accumulate the status of all the suites.

kaheichan:
Hi @venkat1077 
I discussed with @bjrr that we should keep the scope of this ticket and please raise a new ticket for the issue you found
For this PR, I created a small manual test attached in the internal ticket and verified that the summary is getting generated without duplicate entry

venkat1077:
Thanks @kaheichan @bjrr 
Creates WAR-1694 for the above mentioned issue.

PR #92, opened 2017-07-25
WAR-1296:update netconf related files

YutakaMizugaki:
WAR-1296: supports netconf protocol version 1.1
and add get-schema and cancel-commit
WAR-1137: add clear notification buffer
WAR-1138: add print notification buffer

PR #336, opened 2017-12-28
WAR-1680 : Using context as negative in connect_netconf keyword is giving EXCEPTION error

venkat1077:
Issue: Due to the changes in #92, netconf.connect returns None when authentication fails during connection establishment and this needs to be handled in the calling method(s).
Fix: Return False if connection fails in the 'open' method of 'netconf_utils_class.py'

How to test: Run any netconf tests(conenct_netconf kw is enough) with wrong password in the data file. connect_netconf kw was throwing exception earlier, now it is handled correctly and the kw status will be set to False.

### DIFF
--- a/warrior/Framework/ClassUtils/netconf_utils_class.py
+++ b/warrior/Framework/ClassUtils/netconf_utils_class.py
@@ -66,6 +66,9 @@ class WNetConf(object):
                                               nc_session['password'],
                                               nc_session['hostkey_verify'],
                                               nc_session['protocol_version'])
+            # If connection fails return False
+            if self.nc_manager is None:
+                return False
         except:
             traceback.print_exc()
             return False


brettleehari:
Only Ka's review suffice.

warrior/Framework/ClassUtils/netconf_utils_class.py
```diff
@@ -66,6 +66,9 @@ def open(self, session_kwds):
                                               nc_session['password'],
                                               nc_session['hostkey_verify'],
                                               nc_session['protocol_version'])
+            # If connection fails return False
+            if self.nc_manager is None:
```
kaheichan: Hi @venkat1077 
I believe using status = self.nc_manager.isOpen here
and let the except block catch the exception and return false will be a cleaner approach
In the case where netconf.connect changes to return False, this approach may work better

venkat1077: I agree @kaheichan, thanks for the suggestion.
Reasons for keeping this way (i) Most of the known exceptions(during socket/transport creation) are handled and the error_messages and/or tracebacks are printed inside netconf.connect class_method(class_name:client). Since netconf.connect function returns 'None' only when any of these known exceptions occur, I decided not to raise the exception one more time. (ii) To maintain the format of the return statement(last), common across all the methods of netconf_utils_class.py.

kaheichan:
Code logic is fine, Checked internal repo for user reported pass log

PR #339, opened 2018-01-04
War 1599: rewrite of katana class 

keenan35i:
the katana class now works as a pass through meaning you can use any feature from manage.py just instead calling mange.py you call katana.py, added one special case, if the keyword database is after katana.py it will initialize a database of users. currently though if you did that you would get an error because the code for the database is not in apps-demo, however it will be merged soon. other then that use it like you would use manage.py


katana/katana.py
```diff
@@ -14,52 +14,37 @@
...
+
+    def to_string(self, args):
+        args[0] = 'manage.py'
+        if args[1] == 'database':
+            args.remove('database')
+        return 'python ' + ' '.join(args)
+
+    def to_katana_dir(self):
+        dir_path = os.path.dirname(os.path.realpath(__file__))
+        os.chdir(dir_path)
```
sanikakulkarni: Hi @keenan35i,

Please remove line 34, changing directory is not needed.

Thanks.

katana/katana.py
```diff
@@ -14,52 +14,37 @@
 
 import os
 import sys
+import subprocess
+class Katana:
+    def __init__(self):
+        pass
+
+    def runProcess(self, osString):
+        subprocess.Popen([osString], shell=True,
```
sanikakulkarni: Hi @keenan35i ,

Is there a reason why the argument ```shell``` is being set to ```True```?

Thanks.

katana/katana.py
```diff
@@ -14,52 +14,37 @@
...
+
+    def katana_init(self, args):
+        args[0] = 'manage.py'
+        self.to_katana_dir()
+        has_db = True if args[1] == 'database' else False
+        self.runProcess( self.to_string(args) )
+        if has_db:
+            args = self.database_init(args)
+
+    def database_init(self, args):
```
sanikakulkarni: Hi @keenan35i,

I may have forgotten your explanation as to why this database_init function exists. Can you please (re)explain it's usage?

Thanks.

sanikakulkarni:
Hi @keenan35i,

Please change the filename to Katana.py (capital K).

Thanks.

katana/Katana.py
```diff
@@ -0,0 +1,69 @@
...
+        proc.wait()
+        proc = self.runProcess('python manage.py migrate --run-syncdb')
+        proc.wait()
+
+if __name__ == "__main__":
+    if not use_py_server:
+        katana = Katana()
+        katana.katana_init(sys.argv)
+    else:
+        main(port)
```
sanikakulkarni: Hi @keenan35i,

The variable ```port``` on line 69  is not defined - and that throws an error if we ever go into the else block - which we will if django is not installed on the system.

Please evalaute the sys.srgv to get the port number - if it is not given by the user, we can default port number to 8000.

Thanks.

sanikakulkarni:
Hi @keenan35i,

Please run pylint on Katana.py.
You can get pylint with this command: ```sudo apt-get pylint```
To run Katana.py against pylint, navigate to katana directory and run the command: ```pylint Katana.py```

Pylint will list issues with the file formatting - which you will be able to then correct.

Thanks.

PR #340, opened 2018-01-05
WAR-1097: Support for * in path to loop through all project and suite files under directory

venkat1077:
Add support to execute the warrior case/suite xmls matching the pattern(with asterisk) given in the path.
Now asterisk(*) is supported in (i) case_path of a suite & (ii) suite_path of a project

Execute 'wftests/warrior_tests/projects/pj_glob.xml' to test this feature at both suite and case level
Regression and test logs are in Jira ticket


kaheichan:
Hi @venkat1077 
In order to fully test the functionality, I'll build a testsuite like this
TS:
TC1-tc/sample_tc/*.xml
inside the dir, has 3 tc with 1 step - ci regression action-increase value (counter)
TC2-verify_tc-verify the counter value as 3

can apply the same logic to PJ-TS test

venkat1077:
Thanks @kaheichan 
I have added tests to verify the fix as suggested.
Execute 'wftests/warrior_tests/projects/pj_glob.xml' & it will verify this feature at both suite and case level.

kaheichan:
Put the test in Travis

PR #358, opened 2018-01-22
bugfix/WAR-1711: changing the attribute from type to wtype for wtag r…

navvrnd:
…elated attributes in data file. Ran wtag, kw_arg and full regression tests and attached the log in the ticket.


kaheichan:
Verified the code and uses selenium test that failed previously to verify the test passing

PR #371, opened 2018-02-06
WAR-1732 (develop) : RMT when sent via command line does not work as expected

sanikakulkarni:


kaheichan:
Verified that -RMT option works in cli arg
it creats the custom TS and run multiple times as input

PR #372, opened 2018-02-07
WAR-1687 (develop) : Execute multiple test cases via command line by specifying input data file - bugfix

sanikakulkarni:
Use the following command (in 1 line) to test
`python warrior/Warrior -datafile wftests/warrior_tests/data/rest_functional_tests/rest_verify_response_data.xml wftests/warrior_tests/testcases/rest_functional_tests/tc_rest_verify_content_response.xml wftests/warrior_tests/testcases/rest_functional_tests/tc_rest_verify_response.xml`

### DIFF
--- a/warrior/Warrior
+++ b/warrior/Warrior
@@ -166,6 +166,7 @@ def main(parameter_list, a_defects, cse_execution, iron_claw, jiraproj, overwrit
             db_obj.close_connection()
     return status
 
+
 if __name__ == '__main__':
 
     FILEPATH, AUTO_DEFECTS, version, CSE_EXEC, IRON_CLAW, JIRAPROJ, \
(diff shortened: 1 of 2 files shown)


warrior/Warrior
```diff
@@ -114,9 +115,7 @@ def main(parameter_list, a_defects, cse_execution, iron_claw, jiraproj, overwrit
                 print_info('Absolute path: {0}'.format(abs_filepath))
                 if Utils.file_Utils.fileExists(abs_filepath):
                     if len(overwrite.items()) > 0:
-                        default_repo = overwrite
```
kaheichan: The reason why the original code won't work is because
default repo and overwrite both to the same dict which eventually become data_repo
when data_repo is updated during the first execution
its content isn't cleared when executing the 2nd test file from cmd line

kaheichan:
Verified the fix with the command in description in 2 scenarios:
1. testcase file has correct wdf in it
2. testcase file 1 has incorrect wdf, testcase file 2 has empty wdf path
Both situations pass without problem

The code is good, waiting for py3 PR is be created

PR #375, opened 2018-02-09
WAR-1725, WAR-1734(develop) - verify is not working properly in iterative parallel 

venkat1077:
This issue(WAR-1725) not just happens for iterative parallel execution, it happens in all cases when the list iteration is used. Both WAR-1725 and WAR-1734 were working well till 3.6.0 and were introduced by #313 in 3.7.0.

Solution:
1. Took out the fix given in #313(WAR-1445)
2. Given new fix for WAR-1445

Fix explanation:
We use location/index to rearrange the command_list that was iterated using '+' for per_td_block(with yes) iteration. Here I have added a similar logic for list iteration so that the commands derived from list will be treated same like the commands derived using '+'.

Note: WAR-1734 is also resolved since the PR which introduced it, is completely removed here.

Regression logs and tests instructions are in WAR-1725


kaheichan:
Removing myself as main part of the logic is in td iteration

warrior/Framework/Utils/data_Utils.py
```diff
@@ -420,13 +421,6 @@ def get_command_details_from_testdata(testdatafile, varconfigfile=None, **attr):
 
             td_iter_obj = TestDataIterations()
             details_dict, cmd_loc_list = td_iter_obj.resolve_iteration_patterns(details_dict)
-            iter_type = testdata.get('iter_type', None)
-            # Type-2 iteration - per_td_block
-            if iter_type == "per_td_block":
-                details_dict, cmd_loc_list = td_iter_obj.repeat_per_td_block(
-                                                details_dict, cmd_loc_list)
-                details_dict = td_iter_obj.arrange_per_td_block(details_dict,
-                                                                cmd_loc_list)
```
satyakrish: Hi Venkat,
The old logic (lines 423-429 .. now deleted ) was to resolve iteration patterns ('+') and then do list substitutions, this is because iterations should happen first and then the lists should be resolved. The list substitutions do not recognize symbol for iteration ('+') in the command. so if there is a mix of '+' and list substitutions the new logic might fail because we are trying to substitute lists before resolving the '+' symbols.

We are lacking regression tests (old or new) that cover the case where users use iteration and list in the same command . Also there is no documentation in the "warrior/Warriorspace/Config_files/Samples/CLI_Test_Data_Sample.xml" describing the expected behavior for such a scenario.


Can you please add regression tests and documentation for the following cases and make sure the fix covers WAR-1725, WAR-1445, WAR-1734.


Case-1:
iteration_type = per_cmd (default)
cmd_type=1: some commands in the beginning of test data block has no iteration, no lists
cmd_type=2: some commands in the test data block should have only iteration
cmd_type=3: some commands in the test data block should have only lists
cmd_type=4: some commands in test data block should have iteration+lists
cmd_type=5: some commands in the end of test data block should have no iteration, no lists.

Case-2:
Repeat the above for iteration_type= per_td_block 
Do not have repeat=yes on any commands

Case-3:
Repeat the above for iteration_type= per_td_block 
for each command type have two variations one should have repeat=yes and the other should have repeat=no

venkat1077: Hi @satyakrish 

Actually I have not changed the order, even now Warrior will resolve the iteration patterns first and then it will substitute the list values. I have only made the following changes:

(i) Added a code block to treat 'substituted list values' same like "+" iterations by updating the cmd_loc_list(which was not there earlier), this is required for 'per_td_block:yes repetitions'.
(ii) Moved the code block which performs 'per_td_block : yes repetitions" to happen after list substitutions so that it will include list values.

There is no change in the behaviour of mixing "+" and 'list substitutions' in a single command, below is the behaviour:

Use-case: command with mix of '+' & list substitutions
cmd = "<list part> <'+' part>", here <list part> will give 1,2 and <'+' part> will give a,b.

Since we perform '+' iterations first, cmd list will become ["<list part> a", "<list part> b"]. After list substitutions, cmd list will become ["1 a", "2 a", "1 b", "2 b"].

As you mentioned, this behaviour is not documented yet. Please confirm if this is expected, I will add documentation part and I will also add the regression tests.

satyakrish: Thanks, verified that you have not modified the flow.
Yes the behavior looks good, please document it.
Please add regression tests as required to make sure wider coverage and no breakage on other areas.

venkat1077: I have added the suggested regression tests and also documented the behavior in cli td sample file.

PR #384, opened 2018-02-14
WAR-1687 (katana-apps-demo) : Execute multiple test cases via command line by specifying input data file - bugfix

sanikakulkarni:


kaheichan:
The reason for this change is that
`default_repo = overwrite`
is actually pointing to the same dict everytime in a new testcase as default repo
so after 1st tc finished, overwrite/default_repo/data_repo for tc 1 will contain all the data generate in tc1 and get passed as default repo info to tc 2, which is not a desired behavior

The fix is to make a fresh copy of default repo everytime for each tc in the parameters list

The change should go in after katana-apps-demo and develop has merged since Warrior executable has drastically changed in katana-apps-demo

PR #386, opened 2018-02-15
WAR:1725, WAR-1734(katana-apps-demo) - verify is not working properly in iterative parallel

venkat1077:
PR for develop branch : #375 

This issue(WAR-1725) not just happens for iterative parallel execution, it happens in all cases when the list iteration is used. Both WAR-1725 and WAR-1734 were working well till 3.6.0 and were introduced by #313 in 3.7.0.

Solution:
1. Took out the fix given in #313(WAR-1445)
2. Given new fix for WAR-1445

Fix explanation:
We use location/index to rearrange the command_list that was iterated using '+' for per_td_block(with yes) iteration. Here I have added a similar logic for list iteration so that the commands derived from list will be treated same like the commands derived using '+'.

Note: WAR-1734 is also resolved since the PR which introduced it, is completely removed here.

Regression logs and tests instructions are in WAR-1725

### DIFF
--- a/warrior/Framework/ClassUtils/testdata_class.py
+++ b/warrior/Framework/ClassUtils/testdata_class.py
@@ -562,8 +562,6 @@ class TestDataIterations(object):
         cmd_loc_list = [0]
         cmd_size = 1
         for i, cmd in enumerate(cmd_list):
-            cmd_list_subs = []
-            is_iter = False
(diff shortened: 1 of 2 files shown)


kaheichan:
Removing myself as main part of the logic is in td iteration

PR #391, opened 2018-02-19
WAR-1730 (develop): Set of commands must be executed when there is a failure 

sanikakulkarni:


sanikakulkarni:
Note: Do not review. Created a pull request to review the pylint job. Will update whenever the PR is ready to be reviewed.

sanikakulkarni:
PR is ready to be reviewed.

wftests/warrior_tests/testcases/framework_tests/execute_resume/one_invoked_execute_resume.xml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" ?>
+<Testcase>
+    <Details>
+        <Name>one_invoked_execute_resume</Name>
+        <Title>This testcase has one execute and resume onError</Title>
```
satyakrish: the test case title says execute and resume onError .. what does this mean? 
instead it is better to say that "the test case has one execute and resume and the invoked step has executetype=invoked" as this is what is being done in the steps

satyakrish:
please add test cases to cover the following scenario to make sure the following works as expected:
1. execute and resume by providing values that goes up  or ex in step 4 provide execute and resume=1,2,3

2. execute and resume with values that go up and down.

3. execute and resume with multiple steps to invoke some of them should be invoked and some of them should be normal steps.

4. execute and resume to test the value of onerror as abort and goto  on the steps being invoked.

warrior/WarriorCore/onerror_driver.py
```diff
@@ -82,44 +82,88 @@ def getErrorHandlingParameters(node, def_on_error_action, def_on_error_value, ex
...
-    print_info("failed: failure action= goto  %s" % value)
-    error_handle['action'] = 'GOTO'
-    error_handle['value'] = value
+def goto(action, value, error_handle, skip_invoked=True):
+    """returns goto_step_num for on_error action = goto """
+    if skip_invoked:
+        print_info("failed: failure action= goto  %s" % value)
+        error_handle['action'] = 'GOTO'
+        error_handle['value'] = value
+    else:
```
satyakrish: lines 127 to 129 are repeated in multiple places, can we make it a function call

warrior/WarriorCore/testcase_steps_execution.py
```diff
@@ -9,239 +10,349 @@
...
+        self.current_triggered_action = False
+        self.step_status = None
+
+    def execute_step(self, current_step_number, go_to_step_number):
+        """
+        This function executes the determined step - step_num (integer index) from the step_list.
+        This function is called either from the while loop (normal execution) in function
+        execute_steps() or from a for loop (invoked execution)
+        """
+        self.current_step_number = current_step_number
```
satyakrish: consider a list of three steps
[s1, s2, s3]

current implementation:
self.current_step  = s1
self.current_step_num = 0 i.e. the index of s1 in the list
self.next_step  =  1 (self.current_step_num + 1) .... the name is  misleading, it is not the next step but a meaningful number to the current step, in the code this is used so that we dont print list indexes like 0 but more meaningful step numbers in the console.

cange needed: add 1 to self.current_step_num and use it. Dont need self.next_step.

warrior/WarriorCore/testcase_steps_execution.py
```diff
@@ -9,239 +10,349 @@
...
+            if self.skip_invoked:
+                print_info("Skipping this step as it is an invoked step.")
+                return self.next_step, self.go_to_step_number, "continue"
+        # when 'onError:goto' value is less than the current step num,
+        # change the next iteration point to goto value
+        elif self.current_triggered_action and int(self.current_triggered_action) < self.next_step:
+            self.next_step = int(self.current_triggered_action) - 1
+        return self.next_step, self.go_to_step_number, "continue"
+
+    def _execute_step_on_goto(self):
```
satyakrish: _execute_step_on_goto actually executes the current step or if the go_to_step_num matches the current step num

for easy to read/locate puposes can we rename this function to match the functionality somthing like execute_current_step

warrior/WarriorCore/onerror_driver.py
sanikakulkarni: Updated.

warrior/WarriorCore/testcase_steps_execution.py
sanikakulkarni: Updated.

warrior/WarriorCore/testcase_steps_execution.py
sanikakulkarni: Updated.

sanikakulkarni:
All comments have been addressed. New regression logs attached to the JIRA ticket.

satyakrish:
please resolve merge conflicts

satyakrish:
The following comments are not addressed.
please add test cases to cover the following scenario to make sure the following works as expected:

execute and resume by providing values that goes up or ex in step 4 provide execute and resume=1,2,3

execute and resume with values that go up and down.

satyakrish:
We have coverage for RMT=1 after execute and resume starts. Please add a step where a the same step has execute and resume and RMT=1

sanikakulkarni:
@satyakrish,
Addressed all comments.

PR #394, opened 2018-02-26
WAR-1694(develop): Project with parallel suite(s) creates duplicate entries in execution summary

venkat1077:
Expectations of Parallel testcases execution,

1. Separate child process will be created for executing each testcase and copy of master junit object will be sent across all processes.
2. Each process will add the result in the corresponding junit object.
3. After the completion of all processes, those junit objects will be collected to fetch all the newly added results.
4. Warrior will add those results into the original master junit object.

Issue: Filtration is not happening at step3, Warrior adds all the existing results into original master junit object.

Solution: Use Testsuite timestamp to filter out the newly added results so that there will not be any duplicate entries.

Test instructions and regression logs are in Jira ticket.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,9 @@ matrix:
       env: INSTALL=yes TESTFILE=./wftests/warrior_tests/projects/pj_parallel_execution_2.xml
       python: 2.7
     - os: linux
+      env: INSTALL=yes TESTFILE=./wftests/warrior_tests/projects/pj_exec_summary.xml
+      python: 2.7
+    - os: linux
       env: INSTALL=yes TESTFILE=./wftests/warrior_tests/testcases/selenium_tests/tc_selenium_headless.xml SELENIUM=yes
       python: 2.7
       addons:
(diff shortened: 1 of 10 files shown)


venkat1077:
Review comment(s) given by @satyakrish in #395 

I do not see coverage for the following case, can you please add regression test for this.

project : should have suites in parallel
suite-1: exec_type = sequential testcases,

- tc-1: kw parallel (data type = custom)
- tc-2: kw sequential (data type = custom)
- tc-3: kw parallel (data type = iterative)
- tc-4: kw sequential (data type = iterative)

suite-2: exec_type = parallel testcases

- tc-1: kw parallel (data type = custom)
- tc-2: kw sequential (data type = custom)
- tc-3: kw parallel (data type = iterative)
- tc-4: kw sequential (data type = iterative)

Reply from @venkat1077
I have added the suggested tests in regression, execution summary here.

kaheichan:
Hi @venkat1077 
Does the pj_exec_summary require manual verification? If so, can we either put it together with other manual verification test internally or automate the test?

PR #395, opened 2018-02-26
WAR-1694(katana-apps-demo): Project with parallel suite(s) creates duplicate entries in execution summary

venkat1077:
PR for develop branch : #394 

Expectations of Parallel testcases execution,

1. Separate child process will be created for executing each testcase and copy of master junit object will be sent across all processes.
2. Each process will add the result in the corresponding junit object.
3. After the completion of all processes, those junit objects will be collected to fetch all the newly added results.
4. Warrior will add those results into the original master junit object.

Issue: Filtration is not happening at step3, Warrior adds all the existing results into original master junit object.

Solution: Use Testsuite timestamp to filter out the newly added results so that there will not be any duplicate entries.

Test instructions and regression logs are in Jira ticket.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,9 @@ matrix:
       env: INSTALL=yes TESTFILE=./wftests/warrior_tests/projects/pj_parallel_execution_2.xml
       python: 2.7
     - os: linux
+      env: INSTALL=yes TESTFILE=./wftests/warrior_tests/projects/pj_exec_summary.xml
+      python: 2.7
+    - os: linux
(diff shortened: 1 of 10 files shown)


venkat1077:
@satyakrish 
I have added the suggested tests in regression, execution summary [here](https://travis-ci.org/warriorframework/warriorframework/jobs/347214185#L1327).

bjrr:
@venkat1077 We do not need 2 pull requests for war-1694. This is into katana-apps-demo and is not needed. Please update the pr into develop with the comments above and close this pull request.

venkat1077:
Added the comments given here in the develop PR  #394 hence closing this PR.

PR #401, opened 2018-02-28
WAR-1757: dev tools app (stage one)

keenan35i:
this is the first stage of the dev tools app, the first stage allows users to add a clean app template with pre-generated files with the correct structure for warrior platform in mind.


kaheichan:
1. Code needs to be moved to py3 repo
2. App install utils needs to be moved into core, not just copied, import statement in related code also needs to change

PR #406, opened 2018-03-07
WAR-1679: In store_in_repo  keyword user was allowed to update only variable at a time, it should allow to update n number of variable

venkat1077:
Requirement is to support storing multiple values in data_repository using 'store_in_repo' kw by passing a file.
Similar support is already available in set_env_var kw by passing json file, same logic is used here.
Test instructions & Regression results are available in Jira ticket

### DIFF
--- a/wftests/warrior_tests/testcases/common_actions_tests/verify_data_test.xml
+++ b/wftests/warrior_tests/testcases/common_actions_tests/verify_data_test.xml
@@ -3,9 +3,9 @@
         <Details>
                 <Name>verify_data_test</Name>
                 <Title>verify_data</Title>
-                <Engineer>suganya</Engineer>
-                <Date>2017-10-30</Date>
-                <Time>13:13</Time>
+                <Engineer>Warrior_user</Engineer>
+                <Date>2018-03-07</Date>
+                <Time>12:08</Time>
                 <State/>
                 <InputDataFile>No_Data</InputDataFile>
                 <Datatype/>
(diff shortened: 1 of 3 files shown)


kaheichan:
Hi @venkat1077 
The code looks good, just have 1 comment regarding the user_defined_tag steps (last 2 steps)
Can you please add a local_data_test step before step 15, set its execute condition to if var7 == "50"
and put var7 = "50" in the user_defined_tag json block?
This way we can fully test if user_defined_tag is correctly load in step 15 and not preloaded before step 15

wftests/warrior_tests/testcases/common_actions_tests/verify_data_test.xml
```diff
@@ -119,49 +107,120 @@
...
-                                <Rule Condition="" Condvalue="" Else="" Elsevalue=""/>
-                        </Execute>
                         <context>negative</context>
                         <impact>impact</impact>
                         <runmode type="Standard" value=""/>
-                        <Iteration_type type="Standard"/>
+                </step>
+                <step Driver="common_driver" Keyword="store_in_repo" TS="10" draft="no">
+                        <Arguments>
+                                <argument name="filepath" value="../../../../warrior/Warriorspace/Config_files/Samples/Set_REPO_Variable_Sample.json"/>
```
kaheichan: the relative path can be shorten by removing "../../warrior/Warriorspace/"

PR #408, opened 2018-03-15
WAR-1760 support regex in snmp verify keyword

kaheichan:
instead of simple text matching, support regex using the re library in python

### DIFF
--- a/warrior/Actions/SnmpActions/common_snmp_actions.py
+++ b/warrior/Actions/SnmpActions/common_snmp_actions.py
@@ -681,7 +681,7 @@ class CommonSnmpActions(object):
                     testcase_Utils.pNote("No SNMP Result Present!", 'error')
         for element in result_list:
             if mib_string:
-                if mib_string in element[0] and snmp_result in element[-1]:
+                if re.search(mib_string, element[0]) and re.search(snmp_result, element[-1]):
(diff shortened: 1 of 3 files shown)


kaheichan:
PR failing because of snmp file score

warrior/Actions/SnmpActions/common_snmp_actions.py
```diff
@@ -681,7 +681,7 @@ def verify_snmp_action(self, system_name, snmp_result, mib_string=None
                     testcase_Utils.pNote("No SNMP Result Present!", 'error')
         for element in result_list:
             if mib_string:
-                if mib_string in element[0] and snmp_result in element[-1]:
+                if re.search(mib_string, element[0]) and re.search(snmp_result, element[-1]):
```
venkat1077: Add this regex search support in below else block as well(line no. 690).

kaheichan:
WAR-1560 is created to improve snmp pylint score

warrior/Actions/SnmpActions/common_snmp_actions.py
kaheichan: Edited docstring and put the regex in else block, WAR-1560 is the ticket for pylint

PR #409, opened 2018-03-15
WAR-1794 Improve Travis job run time

kaheichan:
1. Move all test except selenium from vm to docker container (using sudo: false setting) to decrease test load time
2. Pack projects together in 2 containers to balance setup time and runtime
3. Make branch build runs simple TC only since branch build is only for blackduck scanning (no duplicate test run)

Selenium test has to run on vm as chrome has permission issue with chromedriver and selenium on non-root environment


satyakrish:
corresponding pr in py3 repo has been merged, this one was missed hence merging it now

PR #410, opened 2018-03-19
WAR-1716 : Project execution mail subject reflected as NONE even though the project is passed

ReshmaManohar:
Requirement:
Subject line of the email sent by Warrior will have the execution status, now it is always 'None' for project execution but it has to reflect the correct execution status.

Fix explanation: 
Project status is computed in sequential/parallel suite(s) execution, possible status values are:

	(i) ERROR(string) - for both error and exception
	(ii) True(boolean)
	(iii) False(boolean)
	(iv) RAN(string)

In Suite/Case, the execution status values thus obtained are converted to corresponding strings (PASS/FAIL/ERROR) using dictionary and gets printed without being stored in a variable
in respective methods. But in project, the values after conversion are stored in a variable named project_status which results in reflection of project execution status as 'None' in
email subject. Therefore, removing the assignment which is not used.

Attached the regression logs and instructions for testing in Jira.

Edit - 04.06.2018 by ReshmaManohar

Included the below modifications,
1. Earlier, mockrun status 'RAN' was reflected as 'None' in mail for Case/Suite/Project. Resolved the same 
    by adding a key 'RAN' to resultconverted dict in email_utils.
2. Also mockrun execution result was 'None' for Suite/Project. Resolved it by adding  a key 'RAN' to dict 
    in  report_project_result method (project_driver.py) and report_testsuite_result 
    method (testsuite_utils.py) respectively.

Attached the logs in Jira.


ReshmaManohar:
On running mockrun, observed a few issues related to this ticket. I am fixing those along with this ticket. Reviewers please hold on until I complete it.

ReshmaManohar:
Reviewers please review the code, done with my modifications.

ReshmaManohar:
Observed the below issue while checking Travis in Py3 repo. Hence fixing the same in this repo also.
Issue:
The computation of final status using impact(compute_status_using_impact method in
testcase_utils_class.py) has break statements when encountered an ERROR or FAIL. So in a Step/Case/Suite, if an error occurred after a fail , the final status is still fail because of break statements.

Fix:
1. Introduced a list that appends all the status values computed using impact(list has the status of all
steps in a case similarly a Suite/Project has their corresponding Case(s)/Suite(s) status)
2. Checked for the presence of ERROR or FAIL in the list. If an ERROR is in list, returned status as 'ERROR'.
If no error but FAIL, returned status as Fail. For cases other than 'RAN' its a PASS.

wftests/warrior_tests/suites/framework_tests/ts_framework_tests.xml
```diff
@@ -22,13 +22,13 @@
...
                 </Testcase>
-                <Testcase>
+        <!--       <Testcase>
                         <path>../../testcases/framework_tests/tc_runmode_checks_on_rup_ruf_rmt_with_intervals.xml</path>
                         <context>positive</context>
                         <runtype>sequential_keywords</runtype>
                         <onError action="next"/>
                         <impact>impact</impact>
-                </Testcase>
+                </Testcase> -->
```
venkat1077: remove this testcase if not applicable/available

warrior/WarriorCore/Classes/testcase_utils_class.py
```diff
@@ -508,22 +508,29 @@ def append_result_files(self, dst_resultfile, kw_resultfile_list, dst_root='Test
...
+                    value = 'RAN'
+                elif input_status is False:
+                    value = False
                 elif input_status is True:
-                    status = True 
+                    value = True
+            result.append(value)
+        if 'ERROR' in result:
+            status = 'ERROR'
+        elif 'ERROR' not in result and False in result:
```
venkat1077: We can take out 'ERROR in result' check from here

wftests/warrior_tests/suites/framework_tests/ts_framework_tests.xml
ReshmaManohar: Removed the testcase.

warrior/WarriorCore/Classes/testcase_utils_class.py
ReshmaManohar: Done

PR #419, opened 2018-04-06
WAR-1669 : User confused on the amount of time Warrior waits for a send command by test data

ReshmaManohar:
Requirement:
The users wants to keep track on the amount of time warrior waits(timeout/sleep) for a send command by testdata.

Fix Explanation:
1. Implemented the wait time as a countdown in wait_for_timeout method (datatime_utils.py) and called 
   this method for both timeout (wait_for_timeout in common_actions.py) and sleep 
   (_send_cmd_get_status method  in warrior_cli_class).
2. Used the countdown approach(wait_for_timeout method in datatime_utils) wherever sleep/timeout is 
   performed.

Attached regression logs and instructions for testing in Jira.


ReshmaManohar:
Need some modifications in the code to avoid unnecessary prints.
Reviewers please hold on the review. Will let u know once done.

ReshmaManohar:
@bjrr @satyakrish 
1. Have used "\033[1A\r" in print (wait_for_timeout method in datetime_utils.py) to bring the cursor back 
    to the beginning of the statement. So that, it prints like a countdown
    Refer to http://ascii-table.com/ansi-escape-sequences.php for more details.
2. Though it prints like a countdown in terminal, multiple prints were logged in console log file. So added 
    a new method called 'print_without_logging' with an argument 'logging' in print_utils. This method will 
    log the data in console log file only if the value of 'logging' argument is True. The default value of this 
    argument is set to True.

ReshmaManohar:
Done with my changes.The code is ready for review.

warrior/Framework/ClassUtils/WNetwork/warrior_cli_class.py
```diff
@@ -1550,12 +1550,16 @@ def send_command(self, command, start_prompt, end_prompt,
...
-                            pNote(tmsg3, "debug")
                             tstamp = Utils.datetime_utils.\
                                 get_current_timestamp()
                         cmd_timedout = True
                         status = "ERROR"
+                        seconds = 60
+                        pNote(tmsg3, "debug")
+                        print_info(("Will wait {0} more seconds to get end prompt " + end_prompt)
+                                   .format(seconds))
+                        Utils.datetime_utils.wait_for_timeout(seconds)
```
venkat1077: Adding this line will make the iteration wait for 60 seconds so 'if int(tdelta) >= 60:' will always pass during the first iteration itself which is wrong. We will have to look for the end prompt after first failure.

ReshmaManohar: Made the required changes.

ReshmaManohar:
@satyakrish 
As suggested removed all cursor position adjustment related code.
1. Have included a argument notify_count to wait_for_timeout method which takes the default value as 4.
2. This divides the wait_time based on the value(notify_count) and notify the user on the remaining wait time.

PR #420, opened 2018-04-09
WAR-1718: HTML Result summary attachment is huge - Need to be Zipped and compressed format

safarheen:
Issue:
-----
When a project is executed with the mail settings. It sends mail with huge Result HTML attachments. (5-7 MB).
- Mail is getting delayed for hours after execution.
- Inbox will soon get flooded with huge mails.

Fix:
---

- Introduced a user-driven approach to compress the result .html file and e-mail the zipped version of it.
- The user can enable/disable this option on need basis.

Attaching logs in JIRA

### DIFF
--- a/warrior/Framework/Utils/email_utils.py
+++ b/warrior/Framework/Utils/email_utils.py
@@ -165,6 +165,8 @@ def compose_send_email(exec_type, abs_filepath, logs_dir, results_dir, result,
                 (1) per_execution(default)
                 (2) first_failure
                 (3) every_failure
+        7. compress - specifies if the result has to be compressed and emailed.
+           Supported options: Yes/No(default)
     """
     resultconverted = {"True": "Pass", "False": "Fail", "ERROR": "Error",
                        "EXCEPTION": "Exception"}.get(str(result))
@@ -176,12 +178,11 @@ def compose_send_email(exec_type, abs_filepath, logs_dir, results_dir, result,
     element = ET.parse(warrior_tools_dir)
     setting_elem = element.find("Setting[@name='mail_to']")
     if setting_elem is not None:
-        co = setting_elem.get("compress")
-        if "Yes" in co:
+        compress = setting_elem.get("compress")
+        if "Yes" in compress:
             report_attachment += ".zip"
         else:
             report_attachment += ".html"
-    print "##### report to be pritnted : ", report_attachment
     # Temporary fix - HTML file can not be attached since it will be generated
     # only after the completion of the warrior execution. Creating html result
     # file at runtime will solve this.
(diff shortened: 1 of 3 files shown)


warrior/Framework/Utils/email_utils.py
```diff
@@ -165,14 +165,24 @@ def compose_send_email(exec_type, abs_filepath, logs_dir, results_dir, result,
                 (1) per_execution(default)
                 (2) first_failure
                 (3) every_failure
+        7. compress - specifies if the result has to be compressed and emailed.
+           Supported options: Yes/No(default)
```
venkat1077: Remove this as there is no arg called compress

warrior/Framework/Utils/email_utils.py
```diff
@@ -165,14 +165,24 @@ def compose_send_email(exec_type, abs_filepath, logs_dir, results_dir, result,
...
+               file_Utils.getNameOnly(file_Utils.getFileName(abs_filepath)) 
+    warrior_tools_dir = Tools.__path__[0]+os.sep+'w_settings.xml'
+    element = ET.parse(warrior_tools_dir)
+    setting_elem = element.find("Setting[@name='mail_to']")
+    if setting_elem is not None:
+        compress = setting_elem.get("compress")
+        if "Yes" in compress:
+            report_attachment += ".zip"
+        else:
+            report_attachment += ".html"
```
venkat1077: I suggest to perform this operation inside 'set_params_send_email' to avoid parsing 'w_settings.xml' multiple times in different places. Construction of 'files' dict can be moved to 'get_email_params' method.

compress attribute should be case insensitive & can accept 'y/n' as well. Use equality to check in if condition.

Compression can happen here instead of html_results_class.py since zip is only for sending email and we should still print the path of the html for the user.

warrior/Framework/Utils/email_utils.py
safarheen: Addressed the comments

warrior/Framework/Utils/email_utils.py
safarheen: Addressed the comments.

warrior/Framework/Utils/email_utils.py
```diff
@@ -9,20 +9,19 @@
...
+"""
 # Utility to send email using smtp
 # Import smtplib for the actual sending function
-
 import smtplib
 import os
 from os.path import basename
 from email import encoders
-from email.MIMEBase import MIMEBase
+from email.mime.base import MIMEBase
```
safarheen: Changing the module to mime.base as as the modules have been renamed.
Reference: https://docs.python.org/2/library/email.html

PR #421, opened 2018-04-19
WAR-1675: As a user, I need an option to set pseudo-terminal dimensions in connect keywords

venkat1077:
Add an option to set the dimensions of the pseudo-terminal(created by the pexpect in spawn class) in connect keywords.

Test instructions, regression results and other details are in Jira ticket.


satyakrish:
@venkat1077  code is good and the tests are good too, one minor suggestion is it possible to change the tag name in data file from just <dimensions> to more meaningful name like <stty_dimensions> or <pty_dimensions> which ever is apt ..

PR #422, opened 2018-04-23
WAR-1381: Not getting executed for y number of systems specified in ID file (when ID is passed using CLI command)

safarheen:
Issue:

- Datafile provided in cmd line should take precedence over the file given at the suite level which is not happening for iterative(parallel/sequential) execution.
- In IterativeTestsuite class(iterative_testsuite_class.py) constructor method, we use the datafile given in 
y) constructor method, we use the datafile given in the suite file for getting the list of systems for iteration instead of the one given at cmd line.

Fix:

- First priority is given for data files specified via CLI
- Default datafiles are the ones specified in the test suite globally.
- If no datafiles are specified at CLI or global level, error is thrown.
- At Suite level execution, step-wise datafiles are not considered.

Testcase validation:

- Verified the fix with the sequential suite:
   wftests/warrior_tests/suites/framework_tests/iter_execution/ts_iter_seq.xml
- This suite has 3 systems to create 3 tmp files.
- Make a copy of this data file with only 2 systems and pass the datafile in CLI. 
- The test should run only for 2 systems.
- Same is applicable for parallel suite also.

### DIFF
--- a/warrior/WarriorCore/testsuite_driver.py
+++ b/warrior/WarriorCore/testsuite_driver.py
@@ -471,10 +471,6 @@ def execute_testsuite(testsuite_filepath, data_repository, from_project,
     suite_end_time = Utils.datetime_utils.get_current_timestamp()
     print_info("[{0}] Testsuite execution completed".format(suite_end_time))
     suite_duration = Utils.datetime_utils.get_time_delta(suite_start_time)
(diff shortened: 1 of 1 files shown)


safarheen:
PR is ready for review. Tested the scenario and attached the logs in JIRA: WAR-1381

PR #425, opened 2018-05-23
WAR-1813: Selenium: _make_browser() function got multiple values for keyword argument 'profile_dir'

safarheen:
Issue:
-------
_makebrowser method has profile_dir as well as **kw_args. However, profile_dir value is obtained from kw_args in previous step. Due to this, now there are two profile_dir attributes in __make_browser function call. This results in above failure.
TypeError: _make_browser() got multiple values for keyword argument 'profile_dir'

**Fix:**  Added logic to remove the profile_dir from kwargs, if exists.

Other Issues fixed along with this are:
Issue-1: 
-------
Binary, gecko_path, gecko_log, proxy_ip, proxy_port values given through datafile where not being read in browser_actions.

**Fix:**  Appended the default argument dict with these attributes. Now, the values are being read from the datafile.

Issue-2:
--------
gecko-log specified in datafile was not being read correctly in browser_management file. 

**Fix**: Updated the file to log the driver info in the path specified in datafile.

Issue-3:
--------
If empty/invalid binary path is passed in the datafile,  inappropriate error message is thrown.

**Fix:**  Provided appropriate error message if empty/invalid binary path is passed in the datafile.

Issue-4:
--------
ff_profile does not get set, if proxy_ip/port are None.

**Fix:**  Added else block to set ff_profile as profile_dir, in case proxy_ip/port are None.

Issue-5:
--------
get_firefox_version() keyword has a version check command specific to Linux only. Hence, execution in windows environment fails.

**Fix:** Added platform specific check in  get_firefox_version() keyword, such that, it builds the firefox version command based on the platform used.

travis.yml file update:
-----------------------
For Selenium 3, it makes sense to add tests with latest selenium version, latest geckodriver and a latest stable firefox version.
Hence,  updated the file.

Please find logs attached in the ticket: WAR-1813

### DIFF
--- a/warrior/Actions/SeleniumActions/browser_actions.py
+++ b/warrior/Actions/SeleniumActions/browser_actions.py
@@ -225,8 +225,6 @@ class browser_actions(object):
             else:
                 output_dict[system_name+"_headless"] = True
                 output_dict["headless_display"] = True
-        else:
-            output_dict[system_name+"_headless"] = False
 
         for browser in browser_list:
             arguments = Utils.data_Utils.get_default_ecf_and_et(arguments, self.datafile, browser)
--- a/warrior/Actions/SeleniumActions/browser_actions.py
+++ b/warrior/Actions/SeleniumActions/browser_actions.py
@@ -47,8 +47,9 @@ class browser_actions(object):
         self.elementlocator_obj = elementlocator_actions()
 
     def browser_launch(self, system_name, browser_name="all", type="firefox",
-                       url=None, ip=None, remote=None, element_config_file=None,
-                       element_tag=None, headless_mode=None):
+                       binary=None, gecko_path=None, gecko_log=None, proxy_ip=None,
+                       proxy_port=None, url=None, ip=None, remote=None, 
+                       element_config_file=None, element_tag=None, headless_mode=None):
         """
         The Keyword would launch a browser and Navigate to the url, if provided by the user.
 
--- a/warrior/Actions/SeleniumActions/browser_actions.py
+++ b/warrior/Actions/SeleniumActions/browser_actions.py
@@ -47,7 +47,7 @@ class browser_actions(object):
         self.elementlocator_obj = elementlocator_actions()
 
     def browser_launch(self, system_name, browser_name="all", type="firefox",
-                       url=None, ip=None, remote=None, element_config_file=None, 
+                       url=None, ip=None, remote=None, element_config_file=None,
                        element_tag=None, headless_mode=None):
         """
         The Keyword would launch a browser and Navigate to the url, if provided by the user.
@@ -232,14 +232,14 @@ class browser_actions(object):
             arguments = Utils.data_Utils.get_default_ecf_and_et(arguments, self.datafile, browser)
             browser_optional_arg_keys = {"binary": None, "gecko_path": None, "proxy_ip": None, 
                                          "proxy_port": None, "gecko_log": None}
-            #Adding browser_optional_arg_keys to arguments to get corresponding values from datafile.
+            # Adding browser_optional_arg_keys to arguments to get corresponding values from datafile.
             arguments.update(browser_optional_arg_keys)
             browser_details = selenium_Utils.\
                               get_browser_details(browser, datafile=self.datafile, **arguments)
             if browser_details is not None:
                 # Call utils to launch correct type of browser
                 # Need to pass the binary, gecko_path, proxy_ip, proxy_port, gecko_log 
-                #  if specified in the datafile
+                # if specified in the datafile
                 browser_optional_args = {}
                 for arg in browser_optional_arg_keys:
                     if browser_details.get(arg) is not None:
--- a/warrior/Actions/SeleniumActions/browser_actions.py
+++ b/warrior/Actions/SeleniumActions/browser_actions.py
@@ -47,9 +47,8 @@ class browser_actions(object):
         self.elementlocator_obj = elementlocator_actions()
 
     def browser_launch(self, system_name, browser_name="all", type="firefox",
-                       binary=None, gecko_path=None, gecko_log=None, proxy_ip=None,
-                       proxy_port=None, url=None, ip=None, remote=None, 
-                       element_config_file=None, element_tag=None, headless_mode=None):
+                       url=None, ip=None, remote=None, element_config_file=None, 
+                       element_tag=None, headless_mode=None):
         """
         The Keyword would launch a browser and Navigate to the url, if provided by the user.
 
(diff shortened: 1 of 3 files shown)


warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py
```diff
@@ -538,6 +540,11 @@ def _make_ff(self, webdriver_remote_url, desired_capabilites, profile_dir, **kwa
                 # This is for internal testing needs...some https cert is not secure
                 # And firefox will need to know how to handle it
                 ff_capabilities['acceptInsecureCerts'] = True
+                
+                if not fileExists(str(binary)):
```
venkat1077: We need to handle two things here:
1. When user doesn't give any value for binary tag in IDF - Info_msg: No value given for binary, default firefox will be used
2. When the user provides some path but that doesn't exist - Warning_msg: Given binary path doesn't exist, default firefox will be used.

Pt. 2 is already there, just add Pt. 1.

safarheen: Addressed the comments.

warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py
```diff
@@ -515,15 +516,20 @@ def _make_ff(self, webdriver_remote_url, desired_capabilites, profile_dir, **kwa
...
+        # if firefox_proxy details are not given, set profile_dir
+        # as the ff_profile.
         if proxy_ip is not None and proxy_port is not None:
             ff_profile = self.set_firefox_proxy(profile_dir, proxy_ip, proxy_port)
-
+        else:
+            ff_profile = profile_dir
         log_dir = get_object_from_datarepository("wt_logsdir") if \
-                  kwargs.get("log_path") in [None, False] else kwargs.get("log_path")
+                  gecko_log in [None, False] else gecko_log
```
sanikakulkarni: Hi @safarheen,
Can you please explain the changes on line 532?

Thanks,
Sanika

sanikakulkarni:
Hi @safarheen,

Can you please explain the flow of the code change - that will help me a lot in doing the review right.

Thanks,
Sanika

warrior/Actions/SeleniumActions/browser_actions.py
```diff
@@ -230,13 +230,16 @@ def browser_launch(self, system_name, browser_name="all", type="firefox",
 
         for browser in browser_list:
             arguments = Utils.data_Utils.get_default_ecf_and_et(arguments, self.datafile, browser)
+            browser_optional_arg_keys = {"binary": None, "gecko_path": None, "proxy_ip": None, 
+                                         "proxy_port": None, "gecko_log": None}
+            # Adding browser_optional_arg_keys to arguments to get corresponding values from datafile.
+            arguments.update(browser_optional_arg_keys)
```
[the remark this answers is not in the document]
safarheen: Hi @sanikakulkarni,

 binary, gecko_path, proxy_ip, proxy_port, and gecko_log are optional arguments, which will not be part of argument list initially, the line 232 with update the details of default arguments only(locals).

the values of  binary, gecko_path, proxy_ip, proxy_port, and gecko_log will be taken from the datafile in line 237 and updated into argument list.

safarheen:
Hi @sanikakulkarni,

These code changes ivolve fix for 4 issues.
The flow is as follows:
browser_actions file:
----------------------
Binary, gecko_path, gecko_log, proxy_ip, proxy_port values are considered as optional arguments which were not being read from datafile, as arguments is initialized to hold locals() which would be the default arguments. Hence, made a dict for opt args and updated these to argument list, so that values are fetched from datafile if any.

browser_mgmt.py:
--------------------
1. As kwargs was introduced, the profile_dir value was being passed twice, to fix this, added a changes to remove the profile_dir if present in kwagrs and pass the profile_dir value as such.
2. Firefox_proxy had an issue with setting preferences if proxy_ip and proxy_port are not passed. In order to give ability to set other preferences using profile_dir, added else block.
3. Initially the webdriver did not have a support to pass the logs of geckodriver. So, a default path was used (which is the execution dir).
Recently, the selenium webdriver supports to pass the log path where the geckodriver logs can be stored. (custamizable by the user) which can be passed through the datafile.
gecko_log attribute was added in browser_action file to implement this update. But, it was not replaced with the default log_path in browser_mngmnt file.
With this change, the gecko_log (path) if passed from datafile, will be considered. If not, the default execution dir will be used to save the geckodriver log.
4. To handle proper error message, if the firefox binary is passed empty. added a block to check if the passed binary value is appropriate, if not it would fetch the default firefox binary.

travis.yml:
----------
Added a test to validate that browser versions >47 work with geckodrivers installed.
Also, selenium V3 is used.

warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py
```diff
@@ -539,6 +545,15 @@ def _make_ff(self, webdriver_remote_url, desired_capabilites, profile_dir, **kwa
...
+                        print_warning("Given firefox binary '{}' does not exist, default "
+                                      "firefox will be used for execution.".format(binary))
+                        binary = None
+                else:
+                    print_info("No value given for firefox binary, default "
+                               "firefox will be used for execution.")
+
                 # Force disable marionette, only needs in Selenium 3 with FF ver < 47
                 # Without these lines, selenium may encounter capability not found issue
                 # https://github.com/seleniumhq/selenium/issues/2739
```
safarheen: Hi @sanikakulkarni 

Query:
-------
If we do not support Windows, it would affect the testing of Nagammai's team testing activities as they have been using both Linux and Windows for their test automation.
As per my knowledge, we do not have an OS independent command to get the firefox version.

Suggestion:
-------------
we could provide support, just for the version command update.
the following code would address this issue:

def get_firefox_version(self, binary):
        """
            Use firefox binary to find out firefox version
            before launching firefox in selenium
        """
        import platform
        raw_version=""
                   
        if platform.system() in "Linux":
            if binary in [False, None]:
                binary = "firefox"
            raw_version = check_output([binary, "-v"]) 
        elif platform.system() in "Windows":
            if binary in [False, None]:
                binary= FirefoxBinary._default_windows_location()
            print_info("*** ff binary for windows is : {} ".format(binary))
            command = "%s -v | more" % (binary)    
            raw_version = check_output(command)          
                
        version = False
        try:
            #raw_version = check_output(ff_version)
            print_info("*** Raw_version is: {} ".format(raw_version))
            match = re.search(r"\d+\.\d+", raw_version)
            if match is not None:
                version = LooseVersion(match.group(0))
            else:
                print_info("Cannot parse Firefox version: {}".format(raw_version))
        except CalledProcessError:
            print_error("Cannot find firefox version, will not launch browser")
        return version

warrior/Actions/SeleniumActions/browser_actions.py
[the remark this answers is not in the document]
sanikakulkarni: You are right. This seems correct. Thanks!

warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py
sanikakulkarni: Got it. Thanks!

warrior/Framework/ClassUtils/WSelenium/browser_mgmt.py
sanikakulkarni: Hi @safarheen,

This code looks fine. Please add it to the codebase (py2 and py3. Please note that the code *may* require changes when being ported to py3) and rerun the selenium projects and attach them to the JIRA ticket. Thanks.

safarheen: Hi @sanikakulkarni 

I have added the platform check in the code for both py and py3. attached the tests in JIRA.
Please approve and merge the PRs

sanikakulkarni:
Selenium regression passed. Approving.

PR #435, opened 2018-07-05
WAR-1116: Remove 'cse script execution' option from warriorframework repo

safarheen:
Issue:
------
cse script execution is not supported in warriorframework repo(OSS) as it is internal to FNC, action file has been removed already but the Warrior executable(function call) & parser methods still has some cse code which needs to be removed. cse option should not be listed in warrior help as well.

Fix:
---
Removed all the cse related code from warriorframework.
Removed from help options as well.

Related logs are attached in JIRA: WAR-1116

### DIFF
--- a/warrior/Framework/Utils/data_Utils.py
+++ b/warrior/Framework/Utils/data_Utils.py
@@ -1490,29 +1490,6 @@ def verify_inorder_cmd_response(match_list, verify_list, system, command,
     return status
 
 
-def get_cse_script_args_string(datafile, system_name):
-    """Form the argument string for the CSE script from the arguments
-    provided in datafile"""
-    args_string = False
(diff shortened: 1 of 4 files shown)


warrior/Framework/Utils/data_Utils.py
```diff
@@ -1490,29 +1490,6 @@ def verify_inorder_cmd_response(match_list, verify_list, system, command,
     return status
 
 
-def get_cse_script_args_string(datafile, system_name):
```
venkat1077: @satyakrish 
This PR is to remove CSE script execution related options from WF.
This util method(get_cse_script_args_string) is specific to CSE script execution and it's usage in non-native warrior keywords is very remote. Should we keep it for backward compatibility?

venkat1077: Spoke to Sathya, we agreed to remove this util method.

PR #437, opened 2018-07-20
WAR-542: As a user, I want to know if a step/testcase passed all attempts when that it was set to RUF & RUP

ReshmaManohar:
Requirement:
1. Currently, if a step/testcase passes all the attempts and if it was set to "RUF", no print messages specifying the end of attempts is given out on the console. A message at the point where those attempts finish would be helpful to the person reading the logs later.

Fix Explanation:
1. Included the print statements at the beginning and end of runmode execution as an indication to user that the runmode execution completed all the provided attempts. 

Added the regression logs and instructions for testing in Jira.


ReshmaManohar:
@satyakrish Have added the logs with the updated print statements in the ticket

PR #438, opened 2018-07-24
WAR-1806: A case/suite with invalid filepath makes the suite/project('s) end result status as ERROR even if the case/suite exec_type is 'No'

safarheen:
Issue:
------
A case/suite with invalid filepath makes the suite/project('s) end result status as ERROR even if the case/suite exec_type is 'No'.

Fix:
---
Added a Check to validate the execType along with the filepath check, so that the invalid filepath error doesnt raise.
Updating the title of the testcase only if the path exists, else passing title as None. This overcomes the invalid filepath scenario for execType = No.

Related logs are attached in JIRA: WAR-1806

### DIFF
--- a/warrior/WarriorCore/sequential_testcase_driver.py
+++ b/warrior/WarriorCore/sequential_testcase_driver.py
@@ -117,7 +117,7 @@ def execute_sequential_testcases(testcase_list, suite_repository,
             print_info("testcase attempt: {0}".format(
                                 testcase.find("retry").get("attempt")))
 
-        if Utils.file_Utils.fileExists(tc_path):
+        if Utils.file_Utils.fileExists(tc_path) or action is False:
             tc_name = Utils.file_Utils.getFileName(tc_path)
             testsuite_utils.pSuite_testcase(junit_resultfile, suite_name,
                                             tc_name, time='0')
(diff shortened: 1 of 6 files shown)


warrior/WarriorCore/sequential_testcase_driver.py
```diff
@@ -192,9 +192,12 @@ def execute_sequential_testcases(testcase_list, suite_repository,
                 data_repository['wt_junit_object'].update_attr(
                                 "status", "SKIPPED", "tc", tmp_timestamp)
                 data_repository['testcase_%d_result' % tests] = "SKIP"
-                title = Utils.xml_Utils.getChildTextbyParentTag(
-                                        tc_path, 'Details', 'Title')
-                title = title.strip() if title else "None"
+                if Utils.file_Utils.fileExists(tc_path):
+                    title = Utils.xml_Utils.getChildTextbyParentTag(
+                                            tc_path, 'Details', 'Title')
+                    title = title.strip()
```
venkat1077: use same old logic here to set tile as "None" when title value is empty

wftests/warrior_tests/projects/pj_cond_var.xml
```diff
@@ -98,5 +98,13 @@
 			<onError action="next" value=""/>
 			<impact>impact</impact>
 		</Testsuite>
+		<Testsuite>
```
venkat1077: Add these test blocks to pj/ts_framework_tests

sanikakulkarni:
Code looks good apart from the minor comment. Please update so that it can be approved.

PR #439, opened 2018-07-24
WAR-1708: Automate Selenium testing on Travis CI

safarheen:
Issue:
------
Jenkins server doesn't have the correct environment to build vm with different version of browsers
contacted tools team and it's suggested to either
1. test webui with docker image on remote browser (which is only part of the functionality in warrior)
2. run it on Travis with full-scale browser/driver support

Fix:
---
Added selenium regression suite to travis.

### DIFF
--- a/wftests/warrior_tests/data/selenium_func_tests/data_browser_launch_selenium3.xml
+++ b/wftests/warrior_tests/data/selenium_func_tests/data_browser_launch_selenium3.xml
@@ -2,7 +2,7 @@
 <credentials>
     <system name="ff_browser_check">
         <headless_mode>Yes</headless_mode>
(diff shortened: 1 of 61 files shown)


wftests/warrior_tests/data/selenium_func_tests/data_browser_launch_selenium3.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" ?>
+<credentials>
+    <system name="ff_browser_check">
+        <headless_mode>Yes</headless_mode>
+        <ip>168.127.47.210</ip>
```
venkat1077: This is an internal IP, do not expose it.

PR #444, opened 2018-08-27
WAR-599: As a user, I need a way to state the final step status when running a step with RUP

ReshmaManohar:
Requirement:
User needs a way compute the final status of a Case/Suite/Project in runmode execution.

Fix Explanation:
1. If the runmode type is RMT, if a failure occurs in any of the attempts, the overall status is marked as a FAIL. (existing way)
2. If runmode type is RUP or RUF, a new attribute named "status" is introduced in runmode tag to compute status, which takes in two values namely,
    - "last_instance" >> The overall status is the status of the last instance(last attempt)
    - "expected"  >> For RUP, if there is a PASS , the overall result is PASS. For RUF, if there is a FAIL 
                               attempt, the overall result is PASS as the failure is a expected one in this case.
3. If the status attribute is left empty or not provided in RUP or RUF, it behaves in the existing way(default) i.e, if any one of the attempts FAIL, the overall status is marked as FAIL.

Added the Regression Logs and Instructions for testing in Jira.

### DIFF
--- a/warrior/WarriorCore/common_execution_utils.py
+++ b/warrior/WarriorCore/common_execution_utils.py
@@ -236,39 +236,44 @@ def compute_status(element, status_list, impact_list, status, impact):
     if runmode is None:
         status_list.append(status)
         impact_list.append(impact)
-    elif runmode.upper() == "RMT":
-        status_list.append(status)
-        impact_list.append(impact)
-    elif runmode.upper() == "RUP":
-        if element.find('runmode').get('status') is None or \
-           element.find('runmode').get('status') == "":
+    else:
+        if element.find('runmode').get('status') not in [None, '', 'last_instance', 'expected']:
+            print_warning("Unsupported value for status. Please provide a valid value. "
+                          "Using the Default value for execution")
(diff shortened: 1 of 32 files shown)


warrior/WarriorCore/common_execution_utils.py
```diff
@@ -225,3 +225,50 @@ def get_retry_from_xmlfile(element):
...
+        impact_list.append(impact)
+    elif runmode.upper() == "RMT":
+        status_list.append(status)
+        impact_list.append(impact)
+    elif runmode.upper() == "RUP":
+        if element.find('runmode').get('status') is None or \
+           element.find('runmode').get('status') == "":
+            status_list.append(status)
+            impact_list.append(impact)
+        elif element.find('runmode').get('status') == 'last_instance' or \
```
santhosh-warrior: Input validation for status. In case if status is assigned any other value then last_instance or expected it should be handled correctly.

ReshmaManohar: Done.

PR #451, opened 2018-11-02
war-2018(Need keyword to find the time taken for a particular step to pass in the test case)

apathapa-warrior:
Attached the logs in jira


apathapa-warrior:
Hi Brian,
 
we have fixed two issues in this branch.
1) keywords for getting time stamp and getting time delta
2) connect_ssh keyword initializes the testdata response dict in data_repo every time (commit from Reshma)
other modified files are related to Reshma's commit mentioned above.

PR #460, opened 2018-12-27
WAR-2041 - common setup and common cleanup cli changes

ryadaval-warrior:
Initial commit.

### DIFF
--- a/warrior/WarriorCore/Classes/war_cli_class.py
+++ b/warrior/WarriorCore/Classes/war_cli_class.py
@@ -320,7 +320,7 @@ class WarriorCliClass(object):
         parser.add_argument('-datafile', action='store', nargs='?',
                             help="overwrite the path of datafile in execution "\
                             "ignore the datafile specified in testcase.xml")
-
+        #to accept -testwrapperfile as command line argument
         parser.add_argument('-testwrapperfile', action='store', nargs='?',
                             help="overwrite the path of testwrapperfile in execution "\
                             "ignore the testwrapperfile specified in testcase.xml")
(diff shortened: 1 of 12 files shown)


warrior/WarriorCore/Classes/war_cli_class.py
```diff
@@ -320,7 +320,11 @@ def parser(cls, arglist):
         parser.add_argument('-datafile', action='store', nargs='?',
                             help="overwrite the path of datafile in execution "\
                             "ignore the datafile specified in testcase.xml")
-
+        #to accept -testwrapperfile as command line argument
```
santhosh-warrior: replace testwrapperfile with wrapper file.  When specified from command line it will skip both suite.xml and also testcase.xml. Update the help string.

warrior/WarriorCore/testcase_driver.py
```diff
@@ -323,7 +323,7 @@ def get_system_list(datafile, node_req=False, iter_req=False):
         return system_list
 
 
-def print_testcase_details_to_console(testcase_filepath, data_repository):
+def print_testcase_details_to_console(testcase_filepath, data_repository, steps_tag):
```
santhosh-warrior: Is step_tag should have default value ?

wftests/warrior_tests/wrapper_files/tw_sample.xml
```diff
@@ -0,0 +1,66 @@
+<?xml version="1.0" ?>
```
santhosh-warrior: The directory need to committed under Warriorspace.

ryadaval-warrior: this will be added in WAR-2066

warrior/WarriorCore/testcase_driver.py
ryadaval-warrior: done

warrior/WarriorCore/Classes/war_cli_class.py
ryadaval-warrior: done

PR #464, opened 2019-01-29
fix for war-2032(Warrior-Reports: Column names are not visible completely in html reports generated after test cases run)

apathapa-warrior:


apathapa-warrior:
Hi @bjrr,

Most of the changes of this fix already merged into the develop branch from 2041.

PR #466, opened 2019-02-08
changes of version.txt and release.txt from release to develop

apathapa-warrior:
-PR for changes of release.txt and version.txt from release-warrior-3.11.0 to develop


bjrr:
not needed into develop

PR #470, opened 2019-02-20
fix for war-2001(The restAPI keyword "Verify Response" should support environment support in json file which is provided as input argumentt in verification json file in verify_response.)

apathapa-warrior:
The restAPI keyword "Verify Response" should support environment support in json file which is provided as input argumentt in verification json file in verify_response.

### DIFF
--- a/warrior/Framework/ClassUtils/rest_utils_class.py
+++ b/warrior/Framework/ClassUtils/rest_utils_class.py
@@ -224,17 +224,14 @@ class WRest(object):
                     try:
                         expected_api_response = JSON.load(open(expected_api_response, 'r'))
                         for key, value in expected_api_response.items():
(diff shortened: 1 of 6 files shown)


warrior/Framework/ClassUtils/rest_utils_class.py
```diff
@@ -495,6 +512,11 @@ def update_output_dict(self, system_name, api_response, request_id, status, i):
 
     @staticmethod
     def get_output_response(api_response):
+        """
+        This method
```
santhosh-warrior: Add more details in the doc strings

apathapa-warrior: Sure @santhosh-warrior

PR #474, opened 2019-02-26
Feature/war 2099(Support for Openapi_netconf_notification. Added support to take in multiple xpath for validation in netconf.py file)

apathapa-warrior:


apathapa-warrior:
- Attached the test case in warriror_fnc_tests

warrior/Framework/ClassUtils/netconf.py
```diff
@@ -349,18 +349,24 @@ def __receive(self):
...
                         pNote("Checking notification: "
                               "##{}##".format(notification))
                         match = False
                         xml = etree.fromstring(notification)
-                        temp = xml.xpath(waitstr[0], namespaces=waitstr[1])
-                        if isinstance(temp, bool) and temp:
-                            match = True
-                        elif isinstance(temp, list) and len(temp) > 0:
-                            match = True
+                        #Contains list of xpath
```
santhosh-warrior: Is it backward compatible ?  The assumption is it always a list.

apathapa-warrior: Hi @santhosh-warrior,
It is backward compatible. The value of waitstr is a tuple , from that tuple we are taking the first element i.e (zero index which is xpath). User can pass single xpath or multiple xpaths ..when we are splitting the output is always list .

PR #487, opened 2019-03-27
fix for war-1722(Warhorn isn't checking pip install status correctly and ignore stderr output)

apathapa-warrior:


warhorn/source/utils.py
```diff
@@ -572,8 +572,11 @@ def install_depen(dependency, dependency_name, logfile, print_log_name,
...
-        output = sp_output.stdout.read()
-        print_info(output, logfile, print_log_name)
+
+        output, error = sp_output.communicate()
+        return_code = sp_output.returncode
+        if return_code > 0:
+            print_info(output, logfile, print_log_name, error)
     except IOError:
         counter = 1
         print_error("Warhorn was unable to install " + dependency_name + " because Warhorn "
```
bjrr: Error message incomplete

apathapa-warrior: Hi @bjrr 

we have not modified anything in error message. Here is the complete snippet of the code.
print_error("Warhorn was unable to install " + dependency_name + " because Warhorn "
                    "does not have write permissions. You need to have admin privileges to "
                    "install anything!", logfile, print_log_name)

PR #499, opened 2019-05-31
Feature/war 2172(Katana :Option to abort the Test case execution when the keyword status is Exception)

apathapa-warrior:


bjrr:
This is not needed. Changes to the features will occur.

